PR #22, opened 2017-04-25
Fix link to react-with-styles example

markdalgleish:


jongold:
Thanks @markdalgleish

PR #24, opened 2017-04-26
Suggest `tail -F` instead of `tail -f`

justjake:
from `man tail`:

> The -F option implies the -f option, but tail will also check to see if the file being followed has been renamed or rotated.  The file is closed and reopened when tail detects that the filename being read from has a new inode number.  The -F option is ignored if reading from standard input rather than a file.

Seems like good practice if sometimes the file vanishes and whatnot.

@jongold


jongold:
whaaaaaaaaaaaaaaaaaaat omg! thanks!

PR #55, opened 2017-04-29
Handle string only errors in RedBox.js

### LINKED ISSUE #50: Invalid prop `error` of type `String` supplied to `RedBox`
| Reporting a bug or issue

I opened the `basic-setup` and started playing around it; I tried to change a color and typo'ed it:
```js
Night: '#333:',
```

Something went wrong.

**Expected behavior:**

`RedBox` renders the correct stacktrace

**Observed behavior:**

image

The error fails to render due to supplied a propType error (?)

```
Warning: Failed prop type: Invalid prop `error` of type `String` supplied to `RedBox`, expected instance of `Error`.
    in RedBox
```

**How to reproduce:**

Follow the instructions above

**Sketch version:**

Version 43.2

petetnt:
This PR fixes #50 by handling `string` only errors in `RedBox.js`. If the encountered error is a string, the `RedBox.js` will render the error in a generic way as early as possible.

In #50 the error was raised by a third party module `chroma.js`, which [only throws strings in the errors](https://github.com/gka/chroma.js/blob/d2c6d917df4ba2b87d8a740de116a0656bcbdfd5/src/converter/in/hex2rgb.coffee#L29) (which is sadly relatively common, I guess).

Before:
image

After:
image

Signed-off-by: petetnt <pete.a.nykanen@gmail.com>

### DIFF
--- a/src/components/RedBox.js
+++ b/src/components/RedBox.js
@@ -17,7 +17,7 @@ type StackFrame = {
   functionName?: string,
   source?: string,
   args?: any[],
-  evalOrigin?: StackFrame,
+  evalOrigin?: StackFrame
 };
 
 const styles = {
@@ -41,7 +41,7 @@ const styles = {
 };
 
 const propTypes = {
-  error: PropTypes.instanceOf(Error).isRequired,
+  error: PropTypes.oneOfType([PropTypes.instanceOf(Error), PropTypes.string]).isRequired,
   // filename: PropTypes.string,
   // editorScheme: PropTypes.string,
   // useLines: PropTypes.bool,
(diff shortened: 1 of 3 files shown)


src/components/RedBox.js
```diff
@@ -17,7 +17,7 @@ type StackFrame = {
   functionName?: string,
   source?: string,
   args?: any[],
-  evalOrigin?: StackFrame,
+  evalOrigin?: StackFrame
```
petetnt: Prettier strips the trailing comma from here 🤔

jongold: Looks like eslint doesn't catch dangling commas in types — cc @ljharb

jongold:
Nice work, thanks @petetnt :)

src/components/RedBox.js
ljharb: It wouldn't; eslint (and the airbnb style guide) doesn't support flow. If you're using Flow, you'd need a separate flow-specific eslint config.

PR #59, opened 2017-05-01
Examples basic form

lloydwheeler:
PR for a new example combining DOM elements and React primitives. The idea was to see how easy it'd be for me to create of PoC of a (semi) working web-component and then render its various states to a sketch document.

The main hurdle was (as far as I'm aware) there not being a react-primitives textbox so that's the reasoning behind having a `TextBox` component that has `primitive` and `web` components inside it, with shared styles across the two.

I imagine there's more than a few changes to be made here and there so let me know where I can look to improve things!


examples/basic-form/src/main.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+});
+
+const Page = ({ sessions }: { sessions: Array<Session> }) => (
+  <View>
+    <Text style={typography.Heading}>Basic Form w/ DOM elements and React Primitives</Text>
+    <Register>
+      {sessions.map(session => (
+        <Space key={session.password} h={spacing.Large} v={spacing.Large}>
+          <View style={styles.register}>
```
jongold: I'd extract this whole thing to a new component ~= `<RegistrationForm>` that takes an email & password - then you could reuse it in the web version too

examples/basic-form/src/web.js
```diff
@@ -0,0 +1,62 @@
...
+    <div style={styles.containerStyle}>
+      <h1
+        style={{
+          ...typography.Heading,
+          fontFamily: fontFamily
+        }}>Basic Form w/ DOM elements and React Primitives. Type a password! 👀</h1>
+      <Register>
+        <div style={styles.register}>
+          <Text style={{...styles.heading}}>Register an Account</Text>
+          <TextBox
```
jongold: See my comment on src/main - you shouldn't have to build this form twice :)

jongold: you can use getDefaultProps for this 🤔

examples/basic-form/src/components/TextBox/primitive.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { View, Text, StyleSheet } from 'react-primitives';
+import styles from './style';
+import StrengthMeter from '../StrengthMeter';
+
+const TextBox = ({ label, value, type }: Props) => (
+  <View style={styles.formElement}>
+    <Text style={styles.label}>{label}</Text>
+    <View style={styles.textbox}>{value}</View>
+    {type === "password" && value.length > 0 &&
```
jongold: You could also render `StrengthMeter` directly in your form - it doesn't need to live inside TextBox explicitly —

~=
```js
<View>
  <TextBox />
  <TextBox />
  <StrengthMeter />
  <Button />
</View>
```

Alternatively, if you wanted it to be attached to the TextBox,
```js
<View>
  <TextBox />
  <TextBox>
    <StrengthMeter />
  </TextBox>
  <Button />
</View>
```

& then in TextBox,
```
{ children && props.children }
```

lloydwheeler:
Thanks for taking the time to look through this, I'll take another look with your suggestions and resubmit!

lloydwheeler:
Thanks for the pointers, I've spent some time refactoring the example.

- In particular being able to just use `index.js` and `index.sketch.js` instead of explicitly calling each component made everything much easier. The `Register` component is just defined once now and is imported by both `main.js` and `web.js`.
- `StrengthMeter` has been decoupled from the `TextBox` component and is now rendered with `{children}`. [Commit here](https://github.com/airbnb/react-sketchapp/pull/59/commits/5fb6ad9655bea4424c9d2dddcfa9414e83c29518).

Just a couple of other things:

1. What's the ettiqutte for updating this PR? Should I close this and create another or do the new commits just get checked? 
2. Should I be generating the documentation? The examples in `examples.md` all point to the master branch so I wasn't sure whether to pre-empt a new demo being added.
3. Running eslint still complains about optional props not having a corresponding `defaultProp` declaration although this is consistent with the other examples?
4. Apologies for the lint dump in the final commit, I'd thought it'd been running automatically. Will get that fixed.

Thanks!

ljharb:
Force push to this PR to update it; please don't open a new one :-)

lloydwheeler:
No problem!

Feel free to rename, by the time I got to the end of building the example I felt `basic-form` wasn't really that well suited anyway.

🙏

PR #105, opened 2017-05-24
Fix link in examples to react-with-styles

hawkins:
I noticed when exploring examples that the link to `react-with-styles` example leads to the `react-primitives` example instead of the `react-with-styles` one, so I thought I'd submit a patch to correct the link.

However, couldn't guarantee that docs would build with the change as gitbook was whining about not having a README file and I'm not familiar with gitbook or react-sketchapp's build process. So, I thought I'd just submit a patch regardless since it's such a tiny change. 😅


jongold:
Thanks for raising this @hawkins — looks like it was fixed in #22 but I forgot to rebuild the docs :'(

(for future reference — to edit Gitbooks, just edit the markdown file and it will be updated when a maintainer rebuilds the docs)

hawkins:
Whoops!  Thanks for the tip tho, I'll keep it in mind for next time. 👍

PR #108, opened 2017-06-01
Add Simple-UI example

jongold:
You don't have to call `render` straight away in `onRun` functions — feel free to take time to make UI and use it to set up documents.

This shows the simplest way to make a text field and use its values to render a document.

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -28,3 +28,6 @@
 
 ### [Timeline w/ Airtable](https://github.com/airbnb/react-sketchapp/tree/master/examples/timeline-airtable)
 ![timeline-airtable](https://cloud.githubusercontent.com/assets/21080/25830456/cdc67bf8-3411-11e7-8998-fdef507ab0d2.png)
+
+## [Simple UI](https://github.com/airbnb/react-sketchapp/tree/master/examples/simple-ui)
+![simple-ui](https://cloud.githubusercontent.com/assets/591643/26658949/2d565576-4622-11e7-9350-4b6aac7096b8.png)
--- a/examples/simple-ui/README.md
+++ b/examples/simple-ui/README.md
@@ -1,0 +1,25 @@
+# Simple UI
+
+## How to use
+Download the example or [clone the repo](http://github.com/airbnb/react-sketchapp):
+```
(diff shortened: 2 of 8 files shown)


arahansen:
Where were you able to find documentation on available APIs for building this UI? Does sketch describe this in their docs?

mathieudutour:
The [AppKit](https://developer.apple.com/documentation/appkit) and Foundation macOS APIs are available. You can read more about it here: http://developer.sketchapp.com/guides/cocoascript/

jongold:
@arahansen this example was cobbled together from some other Sketch plugins to show the https://archetypeapp.com/ developers how they might think about a plugin using react-sketchapp, whilst sticking to common Sketch plugin patterns as much as possible.

In production* we inject webviews into Sketch & use familiar React techniques to create the UI (because I prefer using the web to Cocoa, naturally). hint: using React Primitives is cool because you can do things like rendering React Native components to a webview inside Sketch, and then to Sketch itself.

How I found documentation: months of digging through random Sketch plugin source code, bothering @mathieudutour, digging through the old Sketch developer mailing list, begging the InVision & Facebook Design Tools developers to share their secrets. Unfortunately it wasn't a well documented process at all, and I don't have time to write-up what we learned - sorry I can't be more useful :(

* we have a component browser plugin for Sketch. It evolved from [this](https://twitter.com/jongold/status/866854672421634049?lang=en); I don't have an up to date video though :(

arahansen:
Injecting webviews sounds perfect. I think that could be a very useful showcase given the target audience of this lib (web developers).

If I found the documentation for webviews, I'd be happy to contribute an example plugin that utilizes them 😁

mathieudutour:
shameless plug for a package that makes it easier to use webviews: https://github.com/skpm/sketch-module-web-view :)

jaridmargolin:
I really like the concept of this...

I think based on some of the changes we have made lately to allow more flexibility in rendering single pages, I think `react-sketchapp` could be a really convenient API for authoring certain types of sketch plugins (ex: Color Palette Generator)

@mathieudutour thoughts? We should either pull the trigger on getting this finalized/merged, or we should close it out.

mathieudutour:
I'd rather have this (and the redux) examples in other repos so we don't feel too bad about not updating them. We can still link to them in the readme or something but it would make it clear that it's just POCs

Issue #114, opened 2017-06-09
Layout support for Symbol instances


gabrielecirulli:
Great work on supporting Symbols! 

I had a quick try and I'm curious what's the plan for integrating the Symbol support with css-layout so that Symbol instances may act like first-class citizens within a rendered layout?

As far as I understand right now, and from my limited testing, css-layout requires explicit widths and heights to be passed to a Symbol instance or it will render as a 0x0 layer.

Have you considered feeding css-layout the actual raw contents of the symbol master at layout time, with overrides applied, so that it will take full advantage of the constraints, and then replacing this at render time with an instance of the corresponding symbol whose width and height and position are set based on the layout?

I have a few hours free so I will have a look at the code and maybe take a stab at this, but I cannot guarantee anything. If anyone else is working on this, give me a heads-up!

gabrielecirulli:
There are some potential issues with this that I can think of:

 - Performing proper layout requires replacing the symbol instance with the layers it's composed of in its master. This also requires respecting overrides recursively, and unpacking until you get a hierarchy made entirely out of primitives. This is solvable but might be demanding in terms of resources. 
 - The layout constraints available in Sketch symbols do not fully cover what's possible with flexbox. This means that symbol instances whose sizing has been determined by css-layout may have some visual glitches due to the same sizing being applied on a more limited constraint system. As far as I can tell from looking into the file format, symbol instances in Sketch are represented as a single reference to the master plus optional overrides. This means that there's no way to replace the contents of a symbol instance and render a custom layout inside it. This poses a big limitation. Any ideas how to overcome it? I believe that symbol layout may still be useful for small things such as buttons resizing based on content, but overcoming this larger roadblock may make it a better system.

PR #115, opened 2017-06-09
Update Jest to v20

gabrielecirulli:
Updates to Jest 20 by:

 - Updating snapshots where they subtly changed due to changes in the tool
 - Moving the config to jest.config.js

One of the primary reasons I wanted to do this before working on it further is that under certain versions of Jest and Node, `console.log` calls will be muted and this makes testing quite hard. Let me know if you'd like to make any further changes. I'll probably base another branch on this so that I get a better testing flow, and potentially rebase onto the old version if needed.


gabrielecirulli:
It looks like the tests fail. I used Node 8 instead of 6 which travis does. Have you considered hardcoding the version (e.g. in a .node-version file) and/or updating it to v8? I can help do it, should I do it directly in this PR?

gabrielecirulli:
I took the liberty of updating the node version in this PR just to see if it resolves the test issues. Depending on what you prefer, this can still be split into multiple PRs or merged as one.

gabrielecirulli:
Now getting a strange `npm ERR! ELOOP: too many symbolic links encountered, access 'node_modules/npm/node_modules/has-unicode'`.

I don't really know what would cause that. Any clues?

gabrielecirulli:
@jongold any update on whether this is interesting to you? If so I can pursue it further and maybe try my hand at the symbol layout issue in #114

.node-version
```diff
@@ -0,0 +1 @@
+8.1.0
```
ljharb: This file is only used by nodenv; travis-ci uses nvm. Also is v8.1 even out?

Please remove this file.

.travis.yml
```diff
@@ -5,8 +5,7 @@ cache:
 notifications:
   email: false
 node_js:
-  - '7'
-  - '6'
```
ljharb: Please also restore v6; dropping tests is dropping support, and we should be testing on v6 at least until it's not LTS.

__tests__/components/__snapshots__/RedBox.js.snap
```diff
@@ -73,8 +69,9 @@ exports[`<RedBox /> renders simple errors 1`] = `
           "fontFamily": "Monaco",
           "marginTop": 20,
         }
-      }>
-      attemptAsync
+      }
+    >
```
ljharb: All these style changes should go in a separate PR.

package.json
```diff
@@ -5,6 +5,10 @@
   "main": "lib/index.js",
   "license": "MIT",
   "author": "Jon Gold <jon.gold@airbnb.com> (http://jon.gold)",
+  "engines": {
```
ljharb: engines is advisory only, so it doesn't add much value; that said, this should work on node 6 and npm 3.

__tests__/components/__snapshots__/RedBox.js.snap
ljharb: Ah sorry, ignore this, I didn't realize they were snapshots.

.node-version
gabrielecirulli: The current version of Node seems to be 8.1.1, yeah. I'll remove this.

gabrielecirulli:
@ljharb what should we do about package-lock.json?

gabrielecirulli:
Strangely, the build on Node 6 fails with some bad matches in the snapshots, probably due to changes in internal Node call stacks. On Node 8 it fails with `npm ERR! ELOOP: too many symbolic links encountered, access 'node_modules/npm/node_modules/has-unicode'.`, which as I mentioned couldn't find anything on how to solve. Any ideas?

ljharb:
We should add `package-lock=false` to `.npmrc` and gitignore lockfiles, since only apps should have lockfiles. I'll do that in a separate commit now.

gabrielecirulli:
Apologies, I'm abandoning this as I won't be able to dedicate any further time to it.

Issue #44, opened 2017-04-27
How to add text style to some characters in between a Text ?


brijeshb42:
For adding text styles like bold, italic to whole sentences, it works great. But how can I add a style to some characters in between ?

I am trying to achieve this functionality -

```javascript
<Text style={myStyle}>This is a <bold>word</bold> in a sentence.</Text>
```

lelandrichardson:
Hey @brijeshb42,

Great question! You can do this by nesting `Text` components inside of each other:

```jsx
<Text style={myStyle}>This is a <Text style={{ fontWeight: 'bold' }}>word</Text> in a sentence.</Text>
```

Hope that helps.

lelandrichardson:
Note that you can also use React to make these things more expressive:

```jsx
const Bold = ({ children }) => <Text style={{ fontWeight: 'bold' }}>{children}</Text>

<Text style={myStyle}>This is a <Bold>word</Bold> in a sentence.</Text>

```

brijeshb42:
I tried this previously but this is what I got

<img width="460" alt="screen shot 2017-04-29 at 12 38 40 am" src="https://cloud.githubusercontent.com/assets/717550/25543604/de49ea54-2c74-11e7-985d-a758c5779619.png">

lelandrichardson:
@brijeshb42 you're right. This is not the expected behavior and is a bug.  Will see what we can do about fixing this!

lelandrichardson:
cc @jongold i'm surprised we didn't run into this as an issue. Confirmed that it behaves this way though. We just don't have many internal components that take advantage of this.

LincMitch:
I'm trying to think of a workaround for nested <Text> objects that are formatted. Why not use 
flexbox on 3  
`<Text style={myStyle}>This is a </Text><Bold>word</Bold><Text> in a sentence.</Text>`

myStyle:
`flex-direction: row;
justify-content: flex-start;`

jongold:
I think there's a bug with our text positioning that stops ^^ from working, @LincMitch — could you try it + report back?

mathieudutour:
Just tried
```
<Text style={{flexDirection: 'row' }}>
  This is a <Text style={{fontWeight: 'bold'}}>bold word</Text> in a sentence
</Text>
```

and it's working as expected. I'm wondering if `flexDirection: 'row'` shouldn't be the default style for `Text`?

LincMitch:
[...]

PR #116, opened 2017-06-10
fix multiple text / number children

jongold:
Addresses #44 

— with multiple string-like children inside a `<Text>`, rendering is super weird `props => <Text>foo {props.bar} {props.baz}</Text>` etc
— a text object with a single `number` child doesn't render

This is still kinda WIP —  current progress (bc I have to leave for the weekend):
image

### DIFF
--- a/src/buildTree.js
+++ b/src/buildTree.js
@@ -22,10 +22,14 @@ const INHERITABLE_STYLES = [
   'writingDirection',
 ];
 
+const allStringsOrNumbers = xs =>
+  xs.every(x => typeof x === 'string' || typeof x === 'number');
+
+const processChildren = xs => (allStringsOrNumbers(xs) ? [xs.join('')] : xs);
+
 const reactTreeToFlexTree = (node: TreeNode, context: Context): TreeNode => {
   if (typeof node === 'string') {
     const textStyle = context.getInheritedStyles();
-    // string node
     return {
       type: 'text',
       style: {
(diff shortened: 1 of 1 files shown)


jongold:
Here's another example, with `examples/colors`
image

What works:
```js
// raw (rather than stringified) numbers inside <Text>
const Component = () =>
  <View>{2}</View>

// multiple text (or number) children
const Component = () =>
  <Text>
    Check this out! { foo } { 10 } { bar } baz
  </Text>
```

What doesn't work 💀 
```js
// raw numbers inside View
const Component = () =>
  <View>{2}</View>

// nested <Text> objects - renders but doesn't flow inline
const Component = () =>
  <Text> foo <Text>bar</Text></Text>

// multiple <Text> nodes inside a <View> — renders but doesn't flow inline
const Component = () =>
  <View><Text>foo</Text><Text>bar</Text></View>
```

## TODO:
- [x] raw (rather than stringified) numbers inside `<Text>`
- [x] multiple text (or number) children inside Text
- [x] multiple text (or number) children inside View
- [] raw numbers inside View
- [] nested <Text> objects
- [] multiple <Text> nodes inside a <View>

I'm going to press on with this — last two might be good to split into a new issue though

jongold:
Actually looks like raw numbers inside `<View>` works - must have been using the wrong branch.

I can't replicate the text wrapping issue in the first comment - I'm going to blame it on an issue with Airbnb's component library and report it when I reproduce it.

Ready for review pls

PR #118, opened 2017-06-10
Specify files for npm

francoischalifour:
The [npm `files` property](https://docs.npmjs.com/files/package.json#files) isn't specified, which results in downloading all the `react-sketchapp` repo at each `npm install`.

I added the `lib` directory to the files, as I believe this is the only one we need.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,9 @@
     "react": "*",
     "react-test-renderer": "*"
   },
+  "files": [
+    "lib"
+  ],
   "keywords": [
     "sketch",
     "sketchapp",


package.json
```diff
@@ -45,6 +45,9 @@
     "react": "*",
     "react-test-renderer": "*"
   },
+  "files": [
```
ljharb: The "files" property is very dangerous; if you want to exclude things, use .npmignore.

That said, src and tests should absolutely be included - `npm explore foo && npm install && npm test` should always work - so there's not likely to be much worth ignoring.

francoischalifour: Fair enough - everything else is config files. Feel free to close this PR then.

Issue #122, opened 2017-06-13
Redux store


dominikchrastek:
Hello, is it possible to inject redux store to the component, that is rendered to sketch?

jongold:
I'm not sure, I haven't tried it yet! I don't have time to look into it right now, but you should give it a go and let us know what happens — we can add bugs to the issues list to work on (or we'd love contributions) :)

Components won't update (because they're not clickable), but we could get it working at least so that you can reuse your same code across web / native / sketch :)

That said, remember that in idiomatic redux your components should be 'dumb' and all the `connect` stuff should be happening in a container component :)

jongold:
Ignore ^^^ - I couldn't stop thinking about it so I had a look. It seems to work :)

https://github.com/airbnb/react-sketchapp/pull/124

PR #124, opened 2017-06-16
WIP: add redux example

jongold:
Theoretically you shouldn't be `connect`ing Redux stores to dumb components, _but anyway_, this shows that it doesn't break `react-sketchapp`.

#122


jaridmargolin:
This has been hanging around for a while. It is a cool proof of concept, but do we want to take on the burden of maintaining another example?

@mathieudutour

PR #126, opened 2017-06-20
Update README.md

gokhankurt:
There was no clear definition what npm is. As a designer, I learnt it from my developer friend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 </div>
 
 ## Quickstart 🏃‍
-First, make sure you have installed [Sketch](http://sketchapp.com) version 43 **(v44 is broken currently)**, & a recent [npm](https://nodejs.org/en/download/).
+First, make sure you have installed [Sketch](http://sketchapp.com) version 43 **(v44 is broken currently)**, & a recent [npm](https://nodejs.org/en/download/). - npm means Node.js (note for non-developers from a designer.) 
 
 Open a new Sketch file, then in a terminal:
 ```bash


README.md
```diff
@@ -9,7 +9,7 @@
 </div>
 
 ## Quickstart 🏃‍
-First, make sure you have installed [Sketch](http://sketchapp.com) version 43 **(v44 is broken currently)**, & a recent [npm](https://nodejs.org/en/download/).
+First, make sure you have installed [Sketch](http://sketchapp.com) version 43 **(v44 is broken currently)**, & a recent [npm](https://nodejs.org/en/download/). - npm means Node.js (note for non-developers from a designer.)
```
ljharb: The word "npm" here links directly to the node website; this additional sentence seems redundant to me.

gokhankurt:
It is not redundant actually. Some of my designer friends had the same
issue. So that means there is lack information and hyper link is not clear
enough. Anyway, this is just a recommendation from real experience.

ljharb:
@gokhankurt I totally think it's a great idea to have more explanation here - but saying "X means Y" right next to a literal hyperlink from X to Y is always redundant on the web; if someone doesn't understand how links work, I'm not sure how they can be helped by this project.

Do you have a suggestion for alternative text that might be more instructive? (Separately, the "note for" part doesn't seem necessary regardless)

gokhankurt:
Umm nice question. To separeting the sentence would be better. Step by step
explanation is better than a single sentence. Most of the designer are like
dummy end users. So you you can also say "- Make sure you have installed
npm". What do you think?

ljharb:
Since npm should always come with node, perhaps changing it to say "node" is better.

gokhankurt:
let's do it Nodejs then, more common way

ljharb:
(Also, separately - "Most of the designer are like dummy end users" is both insulting, and wildly inaccurate. Most designers I interact with know at least basic JS and HTML and are able to prototype their designs as needed, so it's clearly not an objective thing. Let's try to be considerate.)

PR #130, opened 2017-06-24
do not process image children

### LINKED ISSUE #129: Symbol image override failed
**Expected behavior:**
image override works
**Observed behavior:**
when I test with examples/symbols,  the override for image is not override with new url, but new url shows as text. 
**How to reproduce:**
git clone react-sketchapp
cd examples/symbols
npm install
npm run render
**Sketch version:**
43.2
image
image

ianhook:
An attempt to resolve #129 which was causing problems for me as well.

I say attempt, because while this works it's unclear to me whether or not images should have children which would be affected.

### DIFF
--- a/src/buildTree.js
+++ b/src/buildTree.js
@@ -42,10 +42,9 @@ const reactTreeToFlexTree = (node: TreeNode, context: Context): TreeNode => {
     };
   }
 
-  let children = node.children || [];
-  if (node.type !== 'image') {
-    children = processChildren(children);
-  }
+  const children = Array.isArray(node.children)
+    ? processChildren(node.children)
+    : [];
   const style = node.props.style || {};
 
   let textStyle;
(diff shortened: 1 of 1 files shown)


jongold:
Thanks @ianhook! 🎊 

I dug into it and I think it's a bug that's more generalizable than just images — looks like `node.children` is `null` rather than `undefined` (when rendered from a symbol override. I'm not sure why 😅 ), which meant it skipped the fallback assignment to `[]`.

I thiiiink it might be more future-proof to fix that fallback than provide a special-case for `image` - do you want to try that direction?

ianhook:
I can dig around some more.  I think that it's actually some kind of special null version of a user defined object, which is why I went this route.

ianhook:
Seems like the problem is that it doesn't skip the default `[]` value and `allStringsOrNumbers([])` is always `true`.

Issue #128, opened 2017-06-24
Symbol Instances Go Missing When Rendering Plugin


arahansen:
This issue is the same one that was described here: https://github.com/airbnb/react-sketchapp/pull/53#issuecomment-306266892

**Expected behavior:**
Symbols that have been updated, but their name has not changed should not break references to existing instances of that symbol.

**Observed behavior:**
Making changes to a symbol will cause instantiations of that symbol to go missing.

**How to reproduce:**

1. generate a symbol from react-sketchapp
2. put the symbol that is now in sketch on a page
3. update the component in code
4. npm run render picks up the change and pushes the change to sketch
5. the symbol you put on the page in step 2 now is "missing"

**Sketch version:**
43

**Please attach screenshots, a zip file of your project, and/or a link to your github project**
Example Code:
```
import { View, Text, makeSymbol, injectSymbols } from 'react-sketchapp'
import React from 'react'

const H1Container = () => <View><Text>Some other Text</Text></View>
H1Container.displayName = 'H1'

makeSymbol(H1Container)

export default context => {
  injectSymbols(context)
}
```

PR #131, opened 2017-06-25
Allow for symbols to be updated by keeping symbolId

ianhook:
Addresses issue #128.

I'm not sure I'm sold on putting all the component definitions in the default function, but that seems like the best way to get the context so that nesting can work.

### DIFF
--- a/src/symbol.js
+++ b/src/symbol.js
@@ -85,10 +85,6 @@ const injectSymbols = () => {
     left += symbolMaster.frame.width + 20;
 
     const newLayer = fromSJSONDictionary(symbolMaster);
-    // symbolsPage.insertLayer_afterLayerOrAtEnd(newLayer, null);
-    // if (Object.hasOwnProperty.call(layers, symbolMaster.symbolID)) {
-    //   layers[symbolMaster.symbolID].removeFromParent();
-    // }
     layers[symbolMaster.symbolID] = newLayer;
   });
 
--- a/src/symbol.js
+++ b/src/symbol.js
@@ -85,14 +85,17 @@ const injectSymbols = () => {
     left += symbolMaster.frame.width + 20;
 
     const newLayer = fromSJSONDictionary(symbolMaster);
-    symbolsPage.insertLayer_afterLayerOrAtEnd(newLayer, null);
+    // symbolsPage.insertLayer_afterLayerOrAtEnd(newLayer, null);
     // if (Object.hasOwnProperty.call(layers, symbolMaster.symbolID)) {
     //   layers[symbolMaster.symbolID].removeFromParent();
     // }
     layers[symbolMaster.symbolID] = newLayer;
   });
 
-  // replaceAllLayersWithLayers(Object.keys(layers).map(k => layers[k]), symbolsPage);
+  replaceAllLayersWithLayers(
+    Object.keys(layers).map(k => layers[k]),
+    symbolsPage
+  );
 
   let notSymbolsPage = array.find(p => String(p.name()) !== 'Symbols');
   if (!notSymbolsPage) {
--- a/src/symbol.js
+++ b/src/symbol.js
@@ -5,7 +5,6 @@ import { fromSJSONDictionary, toSJSON } from 'sketchapp-json-plugin';
 import StyleSheet from './stylesheet';
 import { generateID } from './jsonUtils/models';
 import ViewStylePropTypes from './components/ViewStylePropTypes';
-import type { SketchContext } from './types';
 import buildTree from './buildTree';
 import flexToSketchJSON from './flexToSketchJSON';
 
@@ -15,8 +14,10 @@ const nextId = () => ++id; // eslint-disable-line
 const displayName = (Component: React$Component): string =>
   Component.displayName || Component.name || `UnknownSymbol${nextId()}`;
 
-const mastersNameRegistry = {};
+let mastersNameRegistry = null;
 let existingSymbols = null;
+const layers = {};
+const globalContext = context; // eslint-disable-line
 
 const msListToArray = (pageList) => {
   const out = [];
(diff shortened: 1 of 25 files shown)


jongold:
This looks great. Nested Symbols are a pain even without this project, but I think this is a pretty reasonable compromise to keep things working.

Just a thought - if you _weren't_ nesting symbols, could you just do the `makeSymbol` calls in the default function, and keep the component definitions outside of it?

tehOPEologist:
@jongold i'm currently doing that in a test branch of mine, which works with the initial symbol branch, but seems to fail with this PR.  i'd like to be able to just send my symbols to sketch without needing to apply them on an artboard or anything like that.  upon sending them, they would also update the symbols whereever they're placed on the artboard that's open.

ianhook:
@jongold I could probably update the example to get the Red, Blue and Photo component outside the default. 

I was also thinking that it might not be correct to always replace symbols and maybe making context optional is the way to go for that.

tehOPEologist:
if the components are the source of truth, i would think you would want to update the symbols in place, since sketch is just a renderer for the react component

ianhook:
@tehOPEologist You're probably correct about updating being the only thing that people will generally want to do.  Unless there is a strong reason otherwise, I'll hold off on trying to figure that out.

Also in you're previous comment to @jongold you mentioned that you where having trouble with this branch? I'm not sure I understand the problem though. Part of your comment makes me wonder if it would make sense to break the example into two parts, one without the `Document` part which just shows symbol creation, and one with the `Document` part to show how overriding works.

tehOPEologist:
@ianhook i created a simple setup to only inject a symbol into an artboard in a gist here: https://gist.github.com/tehOPEologist/df073f39fd6afb89bcf79b4b41bfb37b

when trying to run npm run render to inject the symbol into sketch, this error occurs:

```
TypeError: undefined is not an object (evaluating 'context.document').
Plugin “my-command”, line 6032:
»   var pages = context.document.pages();
```

this error occurs on the first line inside the getExistingSymbols function.

ianhook:
@tehOPEologist you need to pass the context into `makeSymbol` with this PR.  I put a more detailed answer in the gist.

tehOPEologist:
omg this is exactly what i wanted to see from this plugin!!!  this is so huge, thank you so much!  the last thing i would like to see is the option to keep the overrides on the symbol when updating, but tbh it's not a huge deal at the moment.  the fact that it's updating the symbols in place and keeping context of them is amazing!  thanks again @ianhook !!!

tehOPEologist:
@ianhook would there be a way to pass multiple symbols to makeSymbol?  seems gross to have a bunch of makeSymbol calls before the injectSymbols call.  thoughts?

jongold:
@mathieudutour sounds good to me

mathieudutour:
I managed to do it on skpm. Playing a bit more with before releasing. One thing to note tho: `context` is only global __after__ the exported function is called. It will fail if you try to access it before (like a `makeSymbol(c, context)` outside the exported function for example)

tehOPEologist:
@mathieudutour which exported function are you referencing when you say that context becomes global after it's called?  can you make a simple gist to explain the change you're working on?  thanks!

mathieudutour:
```
import {makeSymbol} from 'react-sketchapp'

makeSymbol(c, context) // context is not defined so it doesn't work

function foo () {
  makeSymbol(c, context) // // context is defined because the function is called in the exported function
}

export default function () {
  makeSymbol(c, context) // context is defined here so it works

  foo() 
}

```

mathieudutour:
actually, let me think a bit more about it, I might have a better solution so that it's always defined

mathieudutour:
Ok I managed to make it a real global. No restriction to use `context` :) Thanks rubber duck 🦆

tehOPEologist:
xD!  always glad to help!  thanks a bunch!  we should definitely add a simple example of this now that it's a true global for people to reference

ianhook:
So there are a couple of problems now. The most notable being that this code was really dependent on `replaceAllLayersWithLayers` doing some magic to keep override data between plugin runs. It seemed like I was able to get sketch to remember things if I deleted the old symbols manually from the symbols page after running the plugin, so there's code in there to replicate that programatically.  It does not appear to work. I don't have any good ideas on for this.

Testing is also broken because of skpm changes, but fixing that seems like a moot point.

**Steps to reproduce**

1. Run the plugin. It require the git master version of skpm.
2. Create a new page. Put a symbol on it and change an override.
3. Re-run the plugin.

At this point the symbol that you put on the new page should have reverted to default values.

tehOPEologist:
@ianhook i've replicated this issue as well, and i'd rather have the overrides break than the symbols break...  if we could have both, that would be 100% the best solution, but i wouldn't want overrides breaking to stop this release and be able to have iterative symbols available...  just my $0.02

mathieudutour:
I tested all the existing examples with the master branch of skpm, that are all working 🤔

ianhook:
@mathieudutour I'm seeing the following error, on both my branch and master:
`iahook@mac (╯°□°)╯ ~/dev/ianhook/react-sketchapp $ npm run flow

> react-sketchapp@0.11.4 flow /Users/iahook/dev/ianhook/react-sketchapp
> flow

Launching Flow server for /Users/iahook/dev/ianhook/react-sketchapp
Spawned flow server (pid=4913)
Logs will go to /private/tmp/flow/zSUserszSiahookzSdevzSianhookzSreact-sketchapp.log
../../tmp/skpm/node_modules/typechecker/es2015/index.js:7
  7: var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol" ? function (obj) { return typeof obj; } : function (obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; };
                                                                              ^^^^^^^^ string literal `symbol`. This value is not a valid `typeof` return value


Found 1 error`

And of course the whole thing is waiting on the next skpm release. 

The real problem is the UX around regenerating symbols and losing all the overrides, which I don't think is skpm related.  I'm not sure that it's going to work very well in my situation, so I'm not sure where to go from here.

tehOPEologist:
i think the overrides should hydrate themselves in place when the symbol updates.  id'ing overridable elements to track present state upon update, update the symbol, then apply current state of overrides.

mathieudutour:
@ianhook that's a flow error, nothing to do with skpm. I think you can safely flow ignore `.*/node_modules/typechecker.*`

tehOPEologist:
does newLayers have the current overrides for each layer?  ifso, maybe hydrate overrides through that?

ianhook:
@mathieudutour That appears to have the same problems as the code from #135

tehOPEologist:
does sketch make the contents of overrides public anywhere that react-sketchapp has access to listen to?

ianhook:
@jongold I think that it might be best to merge this and open a new ticket to deal with the overrides later.  Or table this and make a new PR that fixes symbols for v45.

jongold:
@ianhook what's the status of this? looks like the bug in v45 was fixed by #146 — is this PR still blocked by anything?

ianhook:
@jongold so actually #146 adds another bug where the old layers are not removed. This PR addresses that. It doesn't work exactly the way I want it to, but do fix the problem from #128. I think that sketch will need to be updated again for this to work the way that I want it.

I think it makes sense to ship this.

src/symbol.js
```diff
@@ -5,17 +5,100 @@ import { fromSJSONDictionary, toSJSON } from 'sketchapp-json-plugin';
...
+  });
+  return symbolId;
+};
+
+const injectSymbols = () => {
+  const globalContext = context; // eslint-disable-line
+  const pages = globalContext.document.pages();
+  const array = msListToArray(pages);
+
+  let symbolsPage = array.find(p => String(p.name()) === 'Symbols');
```
bomberstudios: FYI, there's a `symbolsPageOrCreateIfNecessary()` method in MSDocumentData that does exactly what you want (and supports a Symbols page with a name other than 'Symbols' : )

ianhook:
@jongold is there anything you need me to do to get this merged?

src/symbol.js
ianhook: I'd do this, but it's kind of tangential to the change at the moment (I just moved that code) and I just want to get this out.  Next PR?

arahansen:
@ianhook these changes are awesome, and I'm stoked to try them out.

I have a question for you (and @jongold / anyone else that might have insight): how did you go about developing this? I found debugging very difficult, and finding documentation on APIs hard to find. Any tips, so I might be able to contribute more in the future?

ianhook:
@arahansen As far as I can tell the closest thing there is to docs are https://github.com/abynim/Sketch-Headers, though jumping from there to runnable code required looking at code from other open source sketch plugins. Debugging was largely done with `tail -f ~/Library/Logs/com.bohemiancoding.sketch3/Plugin Output.log` and that was not the most straight forward.

Issue #133, opened 2017-06-27
Broken in Sketch 45


jongold:
Sketch 45 handily removes the [`replaceAllLayersWithLayers`](https://github.com/abynim/Sketch-Headers/commit/a4b61ea30023c7b97db467ca1189f5c6788b1cc2#diff-ccc395008bbb97c5fca92e34aa2b4450L249) method that [we need to be able to render](https://github.com/airbnb/react-sketchapp/blob/master/src/render.js#L18).

[Related thread](http://sketchplugins.com/d/277-replacealllayerswithlayers-removed-in-v45) on the Sketch Plugins forum.

Would be real swell if anyone has time to dig into the new headers and see if there's something we can replace it with.

PR #134, opened 2017-06-28
update renderToSketch to support Sketch 45 API change

gdowens:
#133


mathieudutour:
All the examples work without any changes?

donysukardi:
Actually simple `container.addLayers([layer]);` works too! I believe across different versions.

I'm currently looking for a way to remove all layers before adding the new one (simulating the good ol' `replaceAllLayersWithLayers`).

gdowens:
After thinking about the current rendering paradigm in react-sketchapp, I believe @donysukardi approach makes more sense. Closing this PR in favor of #135.

PR #135, opened 2017-06-28
Replace removed replaceAllLayersWithLayers API (Sketch 45) with addLayers

donysukardi:
Fixes https://github.com/airbnb/react-sketchapp/issues/133.

Alternative solution to https://github.com/airbnb/react-sketchapp/pull/134 as this also removes all the existing layers before adding the new layer (essentially emulating `replaceAllLayersWithLayers`).

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -22,11 +22,20 @@ const renderToSketch = (
   const json = flexToSketchJSON(node);
   const layer = fromSJSONDictionary(json);
 
+  if (container.addLayers === undefined) {
+    throw new Error(`
+     React SketchApp cannot render into this layer. You may be trying to render into a layer
+     that does not take children. Try rendering into a LayerGroup, Artboard, or Page.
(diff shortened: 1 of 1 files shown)


gdowens:
Formerly, with the replaceAllLayersWithLayers method, one could theoretically render into any layer. Now, with the API change it is only possible to render into LayerGroups. 

I believe detecting this and throwing a useful error would be improve the user experience. Can see my earlier PR for an example.

jongold:
wooooooo thanks @donysukardi & @gdowens. this works with 43-44 too?

donysukardi:
Tested it against 43-44 and it's working great!

screen shot 2017-06-29 at 10 08 48 am
screen shot 2017-06-29 at 10 07 08 am

Thanks for the suggestion @gdowens! I've added the check to throws meaningful message if the given container does not support children.

PR #143, opened 2017-07-07
Add support for textDecorationLine property

davidseik:
Hi, this adds support for the textDecorationLine property for text.
Tested on ***Sketch 44.1***
Example of usage:
```js
<Text  style={{textDecorationLine: 'underline'}}>
  Hello World
</Text>
```

which outputs this:
screen shot 2017-07-07 at 11 13 54 am

The API supports the following attributes: `none`, `underline`, `line-through`, `double`.

I have some questions though:
* Should I create some tests around this?
* How about the documentation, does this grant an update?
* There was an autoformat on commit, I didn't make some of those changes. is that okay?

### DIFF
--- a/src/buildTree.js
+++ b/src/buildTree.js
@@ -12,13 +12,14 @@ const INHERITABLE_STYLES = [
   'fontSize',
   'fontStyle',
   'fontWeight',
+  'textAlign',
+  'textDecoration',
   'textShadowOffset',
   'textShadowRadius',
   'textShadowColor',
   'textTransform',
   'letterSpacing',
   'lineHeight',
-  'textAlign',
   'writingDirection',
 ];
 
--- a/docs/styling.md
+++ b/docs/styling.md
@@ -78,6 +78,7 @@ Components use CSS styles + flexbox layout.
 | `fontSize` | `number` | ✅ |
 | `fontStyle` | `normal` &#124; `italic` | ✅ |
 | `fontWeight` | `string` | ✅ |
+| `textDecorationLine` | `none` &#124; `underline` &#124; `double` &#124; `line-through`  | ✅ |
 | `textShadowOffset` | `{ width: number, height: number }` | ⛔️ |
 | `textShadowRadius` | `number` | ⛔️ |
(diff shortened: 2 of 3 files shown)


jongold:
Wooo, great work @davidseik :)

- we haven't found a good solution for tests right now, unfortunately. we're hoping that relying on small functions & flow keeps this mostly error free, but we just test in `examples/` 😱 
- Yes! we can add the textDecoration* props to the Styling documentation :)
- That's fine! Prettier is set up to automatically format files; presumably that file just hadn't been touched since we added it :)

Will dig in this afternon

src/buildTree.js
```diff
@@ -12,13 +12,14 @@ const INHERITABLE_STYLES = [
   'fontSize',
   'fontStyle',
   'fontWeight',
+  'textAlign',
+  'textDecoration',
```
jongold: we mirror the React Native APIs as much as possible — can you change this to `textDecorationLine` to match?

davidseik: Hi Jon, I actually wanted to use the exact same API of textDecorationLine but I found this:
https://github.com/airbnb/react-sketchapp/blob/master/src/stylesheet/expandStyle.js#L61

When the style gets passed down as a textStyle this part expands it to be renamed `textDecoration`; I don't know why this is needed. I can go ahead and remove the shortHand here and add the `textDecorationLine` over here so it's propagated like it. Thoughts?

jongold: 🙃 I think maybe it was actually correct the first time - looks like I have your original commit and it's indeed working properly by expanding the textDecorationLine into textDecoration. Not sure what went wrong the first time I reviewed it, sorry!

Documentation update looks good - we want the public interface to be `textDecorationLine`.

davidseik: Got it, then let me change it back.

davidseik: Funny thing is that because of the expansion they both work `textDecoration` & `textDecorationLine`; Thoughts?

davidseik:
@lelandrichardson @BenjaminCorey can you guys check if this can be merged?

PR #158, opened 2017-07-28
Allow naming symbols; change symbol page selector

ianhook:
screen shot 2017-07-28 at 8 55 26 am

Naming allows symbols to be put into groups, as above. Names are still not required and will be pulled from the object as before if not assigned.

@bomberstudios pointed out the provided way to find the Symbols page in my last PR and I've also added that and tested both things in v43.2 44.1 and 45.2

### DIFF
--- a/examples/symbols/src/my-command.js
+++ b/examples/symbols/src/my-command.js
@@ -9,7 +9,7 @@ const RedSquare = () => (
   </View>
 );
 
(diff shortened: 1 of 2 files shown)


jongold:
@ianhook sorry for the late response! is this good to go?

ianhook:
@jongold it's good

PR #159, opened 2017-08-03
fix(box-shadow): Resolve box-shadow spread

weaintplastic:
Instead of hard-coding the box shadow spread property, this fix will pick up a spread if defined and will default to `0`


jongold:
Thanks @weaintplastic!

weaintplastic:
You more than welcome. Thank's for merging @jongold

PR #164, opened 2017-08-05
Add glamorous example

nitin42:
With the release of [glamorous](https://github.com/paypal/glamorous) v4, we have now support for styling primitive React interfaces using [glamorous-primitives](https://github.com/nitin42/glamorous-primitives) which means it can also be used with `react-sketchapp` 🎉


nitin42:
Done! Sorry, I've been quite busy lately (maintaining projects and interviews). should work now 😄

Issue #112, opened 2017-06-08
Add docs for Symbols


jongold:
Here's an [example](https://github.com/airbnb/react-sketchapp/blob/master/examples/symbols/src/my-command.js) to go by — need to document `makeSymbol` and `injectSymbols`

PR #171, opened 2017-09-07
Documentation for symbols (Issue #112)

kristof:
I've added documentation on how to use symbols, this refers to issue #112.

The symbol documentation is divided in two sections, createSymbol and injectSymbols. The documentation includes examples for basic usage, text overrides, image overrides, nested symbols and how to inject the symbols into the symbols page.

Remarks are very welcome and I'm open to do some changes.


kristof:
I think it is, unless anyone else has any comments.

jongold:
:shipit: :)

Issue #173, opened 2017-09-08
Image primitive is broken within Sketch Beta v47 (45292)


kristof:
**Expected behavior:** 
The `Image` primitive should render an image.

**Observed behavior:**
Instead a red box appears with the following message.

<img width="500" alt="screen shot 2017-09-08 at 10 29 07" src="https://user-images.githubusercontent.com/350388/30202971-9ec96444-9480-11e7-81db-e61d71e31880.png">

**How to reproduce:**
* Install Sketch Beta v47
* Create a .skpmrc file in your home directory
* Add the following contents, to make sure react-sketchapp uses the beta release `{ sketchPath: '/Applications/Sketch Beta.app' }`
* Try to run the profile-cards example contained within the examples folder.

**Sketch version:**
Sketch Beta v47

kristof:
With some further investigation it seems they made some changes to MSImageData within v47.

Current Release – https://github.com/abynim/Sketch-Headers/blob/master/Headers/MSImageData.h
```
- (id)initWithData:(id)arg1 sha:(id)arg2;
- (id)initWithImage:(id)arg1 convertColorSpace:(BOOL)arg2;
```

Beta 47 – https://github.com/abynim/Sketch-Headers/blob/betas/Headers/MSImageData.h
```
- (id)initWithImageConvertingColorSpace:(id)arg1;
- (id)initWithImage:(id)arg1;
```

PR #174, opened 2017-09-08
Fix Image primitive in Sketch 47

kristof:
This is a fix for issue #173, it will remain backwards compatible with Sketch 46. 

@jongold how do we handle API changes from the Sketch side? Will we remove the check once 47 is released or will we remain backwards compatible.


jongold:
Thanks @kristof!

PR #179, opened 2017-09-18
Add Editor configuration files

larsonjj:
# Overview

This PR adds `.editorconfig` files to help automate styling/file rules in most editor/IDE applications and to keep things consistent.

# Why?

This project is awesome, but I found myself needing to manually fix styling before contributing and it took a non-zero amount of time. This PR can help reduce that annoyance.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+# http://editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+end_of_line = lf
+insert_final_newline = true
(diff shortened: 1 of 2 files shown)


jongold:
👍  on the .editorconfig, thanks.

I'm unfamiliar with the .prettierrc — we have `prettier-eslint` [set up](https://github.com/airbnb/react-sketchapp/blob/master/package.json#L63-L68); is this complementary / different? It's pretty important to us that this project follows `eslint-config-airbnb` (naturally) :P

larsonjj:
@jongold,

It was complimentray, but after further review, it looks like the `.prettierrc` is mostly unneeded as I found quite a few prettier plugins for editors can handle the custom prettier-eslint configuration correctly. I removed the file to avoid any potential issues 👍

PR #186, opened 2017-09-22
Allow using array of styles in style prop

jacobp100:


ianhook:
I don't like the fact that this would not be compatible with the React Primitive version of View.

Also, I'm not sure why this is required.  Can you give an example of why you'd need this?

mathieudutour:
I'm pretty sure that's how you use `style` in `react-native` (hence in `react-primitives`):

```js
<View style={[styles.button, styles.primary]} />
```

jacobp100:
It already works. I just get a prop type error.

jongold:
Sweet - closing this in favor of #9

mathieudutour:
Hum but that's a PR and #9 is an issue

jacobp100:
We flatten styles here:

https://github.com/airbnb/react-sketchapp/blob/master/src/components/View.js

@jongold You linked an issue, not a PR. This has actual code to fix this!

jongold:
oh wow my bad 😅

PR #188, opened 2017-09-26
Overhaul layout calculation (css-layout -> yoga-layout)

### LINKED ISSUE #11: Fix default line height
It's noticeable here -
image

~Looks like Sketch defaults to `1.2` line-height whereas browsers (and Native? maybe?) default to `1`.~

from [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height)
> **normal**
> Depends on the user agent. Desktop browsers (including Firefox) use a default value of roughly 1.2, depending on the element's font-family.

Line height seems to be correct, I think bounding boxes are being drawn incorrectly

Check this out:
- the two titles have just been rendered with `lineHeight: 22`. the one on the right has had its bounding box adjusted already; the one on the left is as-it-renders.

After double-clicking on an anchor point on `Four Barrel Coffee`, or resizing it horizontally, the bounding box snaps up several pixels, at which point I reflow the text underneath it.

I _think_ the bounding box is initially getting drawn too big, but I'm not sure why 🤔 
kapture 2017-04-01 at 20 57 15

### LINKED ISSUE #15: Handle z-indexes

### LINKED ISSUE #44: How to add text style to some characters in between a Text ?

### LINKED ISSUE #52: Move from css-layout to yoga
Yoga now has an [emscripten target](https://facebook.github.io/yoga/docs/api/javascript/) and it would be good for us to move over to it instead of using css-layout. Yoga is now what React Native is using, and this will give us the highest level of consistency with those platforms.

### LINKED ISSUE #114: Layout support for Symbol instances

### LINKED ISSUE #148: profile-cards has an invalid image uri
👋 Hello! Thanks for contributing. Please use the template that matches your intention  

I am Quang. In "profile-cards" example, Jackie profile image i invalid.

| Requesting a new feature

**Proposal/Feature-request:**

| Reporting a bug or issue

**Expected behavior:**
In sketchapp, it should show Jackie profile image.

**Observed behavior:**
Sketchapp shows nothing, except a RED rectangle.
In main.js, this uri seems invalid
https://pbs.twimg.com/profile_images/756488692135526400/JUCawBiW_400x400.jpg

**How to reproduce:**

**Sketch version:**

image

Thanks,

### LINKED ISSUE #152: Order of elements rendered is inverted in sketch
Hi, while working on a test project to render elements, one of my colleagues noticed that the elements rendered in sketch are in inverted order on how they appear on the code.

Given this code:
code

One would expect to get a container folder with 3 folders `1`, `2` & `3` in order. but you get this:

sketch

Any clue why this happens? This is special important when doing symbols and you want to offer configuration pieces it should be in correct order.

The actual rendered elements are in correct order.

**Sketch version:**
45.2

### LINKED ISSUE #187: flexGrow/Shink/Basis not supported
Currently the documentation says these are supported.

I believe this would be fixed by upgrading to Yoga (#52).

I believe we want to do something like [this](https://github.com/orta/jest-snapshots-svg/blob/master/src/component-to-node.ts).

I'm happy to look at this. Would you be happy for something like the approach like above?

larsonjj:
# Overview

This PR overhauls nearly everything involved with layout calculation. 

It converts from the existing `css-layout` package to `yoga-layout`, preserves component tree order, adds support for `zIndex`, Implements a new nestable `<Text>` rendering engine, and provides sane default dimensions for Symbol instances. This will bring us into parity with React Native (or at least very close). 

Will now be able to use all the properties listed here: http://facebook.github.io/react-native/releases/0.48/docs/layout-props.html

> NOTE: `aspectRatio` and `direction` have been ignored since they platform specific and outside of the flexbox spec

# Z Index Support

`zIndex` style can be used to change the stacking order of your layers. This effectively allows you to put layers in front or behind each other with whatever order you desire

## Example

***Without zIndex***

```jsx
const Swatch = ({ name, hex }) => (
  <View
    name={`Swatch ${name}`}
    style={{
      height: 96,
      width: 96,
      margin: 4,
      backgroundColor: hex,
      padding: 8,
      position: 'absolute',
      top: index * 50}}
  >
    <Text
      name="Swatch Name"
      style={{ color: textColor(hex), fontWeight: 'bold' }}
    >
      {name}
    </Text>
    <Text name="Swatch Hex" style={{ color: textColor(hex) }}>
      {hex}
    </Text>
  </View>
);
```

screen shot 2017-10-02 at 6 33 48 pm

***With zIndex***

```jsx
const Swatch = ({ name, hex, index }) => (
  <View
    name={`Swatch ${name}`}
    style={{
      height: 96,
      width: 96,
      margin: 4,
      backgroundColor: hex,
      padding: 8,
      position: 'absolute',
      top: index * 50,
      zIndex: index
    }}
  >
    <Text
      name="Swatch Name"
      style={{ color: textColor(hex), fontWeight: 'bold' }}
    >
      {name}
    </Text>
    <Text name="Swatch Hex" style={{ color: textColor(hex) }}>
      {hex}
    </Text>
  </View>
);
```

screen shot 2017-10-02 at 6 33 59 pm

Note that the stacking of Swatches has swapped.

# Fix `Missing Fonts` warning

`Helvetica` is now used as the default font if no `fontFamily` is specified on a component. This effectively removes the annoying `Missing Fonts` warning:
screen shot 2017-10-02 at 6 04 21 pm

# New Text Layout Engine

`<Text>` components are now nestable, cascade their styles down to their children, and replicate React Native text layout [specifications](https://facebook.github.io/react-native/docs/text.html#containers).

> NOTE: Only `<Text>` components are allowed to be children of other `<Text>` components. All other components (`<View>`, `<Image>`, etc) will throw an error. This allows for the highest cross-platform compatibility.

## Example

```jsx
class BoldAndBeautiful extends Component {
  render() {
    return (
      <Text style={{fontWeight: 'bold'}}>
        I am bold{' '}
        <Text style={{color: 'red'}}>
          and red
        </Text>
      </Text>
    );
  }
}
```

screen shot 2017-10-01 at 3 11 06 pm

# Changes
- Removed `css-layout` package
- Added `yoga-layout` package
- Tree/layout computation now uses yoga-layout to determine layout properties: width, height, top, bottom, left, and right
  - Symbol instances now use their symbol master dimensions (height/width) by default when rendered. No more 0x0 px issues if no width or height styles are provided.
  - Stacking order is fixed so rendering layers now matches component tree order
- `<Text>` components now render according to the React Native [spec](https://facebook.github.io/react-native/docs/text.html#description)
- Added support for `zIndex` style property
- Set default font to `Helvetica` in order to get rid of `Missing Fonts` message in Sketch
- Travis CI builds now use OSX as it allows for yoga to build correctly.
- Fixed example twitter image errors
- Added some more unit tests

# Breaking Changes

- `<Text>` components no longer accept [View style properties](https://github.com/vhpoet/react-native-styling-cheat-sheet#view) (`backgroundColor`, `borderWidth`, etc). This is due to now allowing Nesting of `<Text>` components and Sketch not supporting any View style properties within Text Layers. Unfortunately, it is just not feasible to to try and dynamically generate `<View>` components that map directly to text characters and locations, which is what would be required to keep this implementation. The good news is that this issue is easy to resolve by just adding a parent `<View>` component to your `<Text>` component and applying view styles to it. (Example: https://github.com/airbnb/react-sketchapp/pull/188/files#diff-d29c4465d14e21519b5b9e53e5b90cbb)
- Layer stacking has been reversed, which makes sure the order you write your components matches the order within Sketch (Fix requested here: #152). This could potentially break some layouts if you were relying on the current stacking order to place layers above/below each other. The solution would be to add `zIndex` style to correct the stacking order, or just reverse the order you wrote your child components.

# Caveats

The existing `yoga-layout` package comes with a built-in asmjs file, but it will fail to work as of `v1.6.0`. This is due to the file not being compiled with the latest emscripten `v1.37.0`. It spits out this error:  

``` 
TypeError: {} is not a function. (In '{}("")', '{}' is an instance of Object) 
``` 

  This means I had to add a hacky script called `yogaFix.js` to the project. All it does, is RegEx replace the `{}("")` code within the asm.js yoga-layout script with `{}`. This effectively fixes the problem and allows for yoga to run correctly within the Sketch JavascriptCore environment.

I'm not a huge fan of this solution, but it's simple and effective. It can also be removed once a newer asm.js version of yoga is released.

# Testing this PR

You can test this PR with the following steps:

- git clone this repo
- checkout the `feature/yoga_layout` branch
- install all packages: `npm install`
- try out any of the example projects within the `examples/` folder

# Relavant Discussions
Closes #187
Closes #52 
Closes #44 
Closes #114 
Closes #152 
Closes #15 
Closes #148 
Closes #11 

# TODOS
- [x] Use yoga to calculate sketch tree
- [x] Test that all examples are working correctly (compared to css-layout)
- [x] Correctly implement Nested Text (https://facebook.github.io/react-native/docs/text.html#description)
- [x] Calculate layout for Sketch symbols
- [x] Preserve component tree order (See #152)
- [x] Support `zIndex`
- [x] Write tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
     "watch": "npm run build:react -- --watch"
   },
   "dependencies": {
-    "css-layout": "^1.1.1",
     "error-stack-parser": "^2.0.0",
     "invariant": "^2.2.2",
     "murmur2js": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -39,7 +39,8 @@
     "prop-types": "^15.5.8",
     "sketch-constants": "^1.0.2",
     "sketchapp-json-flow-types": "^0.2.0",
-    "sketchapp-json-plugin": "^0.0.3"
+    "sketchapp-json-plugin": "^0.0.3",
+    "yoga-layout": "^1.6.0"
   },
   "peerDependencies": {
     "react": "*",
--- a/package.json
+++ b/package.json
@@ -28,7 +28,9 @@
     "test": "jest --config .jestrc",
     "test:ci": "jest --config .jestrc --runInBand",
     "test:watch": "npm run test -- --watch",
-    "watch": "npm run build:react -- --watch"
+    "watch": "npm run build:react -- --watch",
+    "postinstall": "npm run yogafix",
+    "yogafix": "node ./yogafix"
   },
   "dependencies": {
     "error-stack-parser": "^2.0.0",
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "sketch-constants": "^1.0.2",
     "sketchapp-json-flow-types": "^0.2.0",
     "sketchapp-json-plugin": "^0.0.3",
-    "yoga-layout": "^1.6.0"
+    "yoga-layout": "1.6.0"
   },
   "peerDependencies": {
     "react": "*",
--- a/src/jsonUtils/computeNode.js
+++ b/src/jsonUtils/computeNode.js
@@ -421,18 +421,9 @@ const computeNode = (node: TreeNode, context: Context) => {
     }
   }
 
-  // if Text node
-  if (
-    node &&
-    node.type === 'text' &&
-    node.children &&
-    (typeof node.children[0] === 'string' ||
-      typeof node.children[0] === 'number')
-  ) {
-    const content = String(node.children[0]);
+  if (typeof node === 'string' || typeof node === 'number') {
     const textStyle = context.getInheritedStyles();
-
-    yogaNode.setMeasureFunc(createStringMeasurer(node, content, textStyle));
+    yogaNode.setMeasureFunc(createStringMeasurer(node, node, textStyle));
   }
 
   return yogaNode;
--- a/src/jsonUtils/computeNode.js
+++ b/src/jsonUtils/computeNode.js
@@ -3,6 +3,9 @@ import * as yoga from 'yoga-layout';
 import type { TreeNode, ViewStyle } from '../types';
 import Context from '../utils/Context';
 import createStringMeasurer from '../utils/createStringMeasurer';
+import hasAnyDefined from '../utils/hasAnyDefined';
+import pick from '../utils/pick';
+import { INHERITABLE_FONT_STYLES } from '../utils/constants';
 
 // flatten all styles (including nested) into one object
 const getStyles = (node: TreeNode): ViewStyle | Object => {
@@ -421,9 +424,19 @@ const computeNode = (node: TreeNode, context: Context) => {
     }
   }
 
-  if (typeof node === 'string' || typeof node === 'number') {
+  if (typeof node === 'string') {
     const textStyle = context.getInheritedStyles();
-    yogaNode.setMeasureFunc(createStringMeasurer(node, node, textStyle));
+    yogaNode.setMeasureFunc(createStringMeasurer(node, textStyle));
+    return yogaNode;
+  }
+
+  if (
+    node.type === 'text' &&
+    node.props.style &&
+    hasAnyDefined(style, INHERITABLE_FONT_STYLES)
+  ) {
+    const inheritableStyles = pick(style, INHERITABLE_FONT_STYLES);
+    context.addInheritableStyles(inheritableStyles);
   }
 
   return yogaNode;
--- a/src/jsonUtils/computeNode.js
+++ b/src/jsonUtils/computeNode.js
@@ -7,6 +7,32 @@ import hasAnyDefined from '../utils/hasAnyDefined';
 import pick from '../utils/pick';
 import { INHERITABLE_FONT_STYLES } from '../utils/constants';
 
+const walkTextTree = (
+  textTree: TreeNode,
+  context: Context,
+  textNodes: Array<Object>
+) => {
+  if (typeof textTree === 'string') {
+    textNodes.push({
+      textStyles: context.getInheritedStyles(),
+      content: textTree,
+    });
+  }
+
+  if (textTree.children) {
+    for (let index = 0; index < textTree.children.length; index += 1) {
+      const textComponent = textTree.children[index];
+      walkTextTree(textComponent, context.forChildren(), textNodes);
+    }
+  }
+};
+
+const textTreeToNodes = (
+  root: TreeNode,
+  context: Context,
+  textNodes: Array<Object>
+) => walkTextTree(root, context, textNodes);
+
 // flatten all styles (including nested) into one object
 const getStyles = (node: TreeNode): ViewStyle | Object => {
   let style = node.props.style;
@@ -322,7 +348,22 @@ const computeNode = (
 
     // Handle Text Children
     if (node.children && node.children.length > 0) {
-      console.log(node.children);
+      const textNodes = [];
+      const childContext = context.forChildren();
+      // console.log(node.children);
+      for (let index = 0; index < node.children.length; index += 1) {
+        const textNode = node.children[index];
+        if (typeof textNode === 'string') {
+          textNodes.push({
+            content: textNode,
+            textStyles: childContext.getInheritedStyles(),
+          });
+        } else if (textNode.children && textNode.children.length > 0) {
+          console.log(textNode);
+          textTreeToNodes(textNode, childContext, textNodes);
+        }
+      }
+      console.log(textNodes);
     }
 
     return { node: yogaNode, stop: true };
--- a/src/jsonUtils/computeNode.js
+++ b/src/jsonUtils/computeNode.js
@@ -7,6 +7,7 @@ import hasAnyDefined from '../utils/hasAnyDefined';
 import pick from '../utils/pick';
 import computeTextTree from './computeTextTree';
 import { INHERITABLE_FONT_STYLES } from '../utils/constants';
+import { getSymbolMasterByName } from '../symbol';
 
 // flatten all styles (including nested) into one object
 const getStyles = (node: TreeNode): ViewStyle | Object => {
@@ -31,6 +32,14 @@ const computeNode = (
   const yogaNode = yoga.Node.create();
   const hasStyle = node.props && node.props.style;
   const style: ViewStyle | Object = hasStyle ? getStyles(node) : {};
+
+  // Setup default symbol instance dimensions
+  if (node.type === 'symbolinstance') {
+    const symbolProps = node.props;
+    const { frame } = getSymbolMasterByName(symbolProps.masterName);
+    yogaNode.setWidth(frame.width);
+    yogaNode.setHeight(frame.height);
+  }
 
   if (hasStyle) {
     // http://facebook.github.io/react-native/releases/0.48/docs/layout-props.html
--- a/src/jsonUtils/computeNode.js
+++ b/src/jsonUtils/computeNode.js
@@ -22,6 +22,8 @@ const isPercent = (value: number | string) => /%/.test(String(value));
 
 const isAuto = (value: number | string) => String(value) === 'auto';
 
+const isNullOrUndefined = (value: any) => value === null || value === undefined;
+
 const computeNode = (node: TreeNode, context: Context) => {
   const yogaNode = yoga.Node.create();
   const hasStyle = node.props && node.props.style;
@@ -31,7 +33,7 @@ const computeNode = (node: TreeNode, context: Context) => {
     // http://facebook.github.io/react-native/releases/0.48/docs/layout-props.html
 
     // Width
-    if (style.width) {
+    if (!isNullOrUndefined(style.width)) {
       if (isPercent(style.width)) {
         yogaNode.setWidthPercent(style.width);
       } else if (isAuto(style.width)) {
@@ -42,7 +44,7 @@ const computeNode = (node: TreeNode, context: Context) => {
     }
 
     // Height
-    if (style.height) {
+    if (!isNullOrUndefined(style.height)) {
       if (isPercent(style.height)) {
         yogaNode.setHeightPercent(style.height);
       } else if (isAuto(style.height)) {
@@ -53,7 +55,7 @@ const computeNode = (node: TreeNode, context: Context) => {
     }
 
     // Min-Height
-    if (style.minHeight) {
+    if (!isNullOrUndefined(style.minHeight)) {
       if (isPercent(style.minHeight)) {
         yogaNode.setMinHeightPercent(style.minHeight);
       } else {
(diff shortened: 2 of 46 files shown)


package.json
```diff
@@ -28,18 +28,20 @@
...
     "error-stack-parser": "^2.0.0",
     "invariant": "^2.2.2",
     "murmur2js": "^1.0.0",
     "normalize-css-color": "^1.0.1",
     "prop-types": "^15.5.8",
     "sketch-constants": "^1.0.2",
     "sketchapp-json-flow-types": "^0.2.0",
-    "sketchapp-json-plugin": "^0.0.3"
+    "sketchapp-json-plugin": "^0.0.3",
+    "yoga-layout": "^1.6.0"
```
mathieudutour: Probably safer to pin the dependency if we are to postinstall modify it

src/jsonUtils/computeNode.js
```diff
@@ -0,0 +1,457 @@
...
+    }
+    if (style.borderHorizontal) {
+      yogaNode.setBorder(yoga.EDGE_HORIZONTAL, style.borderHorizontal);
+    }
+    if (style.border) {
+      yogaNode.setBorder(yoga.EDGE_ALL, style.border);
+    }
+
+    // Flex
+    if (style.flex) {
```
jacobp100: We need to explicitly check for `undefined` in these, so we don't ignore zero values

jacobp100:
I manually built yoga, and it worked on its own. Maybe we could publish something like `yoga-layout-prebuilt`?

src/jsonUtils/computeNode.js
```diff
@@ -0,0 +1,457 @@
...
+    if (!style.height) {
+      yogaNode.setHeight(measure.height);
+    }
+
+    // Use hardcoded width if present
+    if (style.width || typeof style.width === 'number') {
+      return yogaNode;
+    }
+
+    yogaNode.setWidth(measure.width);
```
jacobp100: You can use `node.setMeasureFunc` for this

larsonjj: Is there any extra advantage to using `node.setMeasureFunc`?

src/jsonUtils/computeNode.js
larsonjj: Good call. Fixed

package.json
larsonjj: Good call. Fixed.

src/jsonUtils/computeNode.js
jacobp100: You’ll need it to get multi-line text working. Especially true when flex is involved, as you don’t know the width at this point.

larsonjj: Fair enough. Done.

Seems like there might be a bit more work to be done within `createStringMeasurer.js` as multiline text doesn't seem to work correctly, but we're on the right track.

src/jsonUtils/computeNode.js
```diff
@@ -0,0 +1,445 @@
...
+        yogaNode.setDisplay(yoga.OVERFLOW_SCROLL);
+      }
+      if (style.overflow === 'hidden') {
+        yogaNode.setDisplay(yoga.OVERFLOW_HIDDEN);
+      }
+    }
+
+    // Flex direction
+    if (style.flexDirection) {
+      if (style.flexDirection === 'row') {
```
jacobp100: Would it be worth extracting these into an object of `flexFirectionModes = { row: yoga.FLEX_DIRECTION_ROW ... }`?

src/jsonUtils/computeNode.js
```diff
@@ -0,0 +1,445 @@
...
+      }
+    }
+    if (!isNullOrUndefined(style.paddingLeft)) {
+      if (isPercent(style.paddingLeft)) {
+        yogaNode.setPaddingPercent(yoga.EDGE_LEFT, style.paddingLeft);
+      } else {
+        yogaNode.setPadding(yoga.EDGE_LEFT, style.paddingLeft);
+      }
+    }
+    if (!isNullOrUndefined(style.paddingRight)) {
```
jacobp100: Would it make sense to extract these `setPadding` into a function that takes the edge and a value to remove duplication?

src/jsonUtils/computeNode.js
```diff
@@ -0,0 +1,445 @@
...
+      if (isPercent(style.minHeight)) {
+        yogaNode.setMinHeightPercent(style.minHeight);
+      } else {
+        yogaNode.setMinHeight(style.minHeight);
+      }
+    }
+
+    // Min-Width
+    if (!isNullOrUndefined(style.minWidth)) {
+      if (isPercent(style.minWidth)) {
```
jacobp100: Apparently this is handled [here](https://github.com/facebook/yoga/blob/master/javascript/sources/entry-common.js#L154-L155)

larsonjj: Nice catch! Updated to remove that extra code

src/jsonUtils/computeNode.js
larsonjj: For now, I prefer being a bit more verbose here to make the code more easily understandable. I messed with trying to dynamically set properties using maps, but it made the code much harder to figure out :( . Definitely open to ideas, but maybe there can be a separate PR to address the verboseness?

src/jsonUtils/computeNode.js
larsonjj: See my previous comment.

src/jsonUtils/computeNode.js
jacobp100: 👍

src/utils/createStringMeasurer.js
```diff
@@ -1,51 +1,39 @@
...
-      );
-
-    // TODO(lmr): handle other widthModes, and height/heightModes
-
-    return {
-      width: 0 + rect.size.width,
-      height: 0 + rect.size.height,
-    };
-  };
+const createStringMeasurer = (textNodes: TextNodes) => (
```
jacobp100: So glad you fixed this! 🎉

larsonjj: 👍

larsonjj:
Hey all,

I finally think I have this PR in a solid spot. Would love to get some feedback/testing on real-world projects if possible. All of the examples seem to work well and mirror their `css-layout` counterparts, so that's a great start. This code enables a lot of awesome stuff, but it is a complete overhaul of the layout engine for this project, so I want to make sure I've accounted for as much as possible.

Any help is greatly appreciated 👍

src/jsonUtils/computeYogaNode.js
```diff
@@ -0,0 +1,336 @@
...
+    if (!isNullOrUndefined(style.marginRight)) {
+      yogaNode.setMargin(yoga.EDGE_RIGHT, style.marginRight);
+    }
+    if (!isNullOrUndefined(style.marginVertical)) {
+      yogaNode.setMargin(yoga.EDGE_VERTICAL, style.marginVertical);
+    }
+    if (!isNullOrUndefined(style.marginHorizontal)) {
+      yogaNode.setMargin(yoga.EDGE_HORIZONTAL, style.marginHorizontal);
+    }
+    if (!isNullOrUndefined(style.margin)) {
```
jacobp100: I'm almost certain you're already correct, but can we check that the order isn't important? I.e. `{ marginTop: 5, margin: 10 }` should have a top margin of 5px.

src/jsonUtils/computeYogaNode.js
```diff
@@ -0,0 +1,336 @@
...
+    }
+    if (!isNullOrUndefined(style.paddingHorizontal)) {
+      yogaNode.setPadding(yoga.EDGE_HORIZONTAL, style.paddingHorizontal);
+    }
+    if (!isNullOrUndefined(style.padding)) {
+      yogaNode.setPadding(yoga.EDGE_ALL, style.padding);
+    }
+
+    // Border
+    if (!isNullOrUndefined(style.borderTop)) {
```
jacobp100: I think these are actually `borderTopWidth` etc. And apparently, there's no support for `borderHorizontalWidth` or `borderVerticalWidth` (but margin and padding *do* have these). Go figure!

src/jsonUtils/computeYogaNode.js
larsonjj: Order shouldn't matter here as I don't call `calculateLayout()` on any of the nodes, until all of their layout styles have already been populated.

jacobp100: Didn't think it would matter. Just double checking!

src/jsonUtils/computeYogaNode.js
larsonjj: Hmm..weird. I'm going off of the Node.h file from the yoga JS source (https://github.com/facebook/yoga/blob/master/javascript/sources/Node.hh#L102). Seems as though there is only `setBorder(int edge, double border);` available. Is there better information on this elsewhere?

jacobp100: I meant the styles---the yoga stuff is fine. http://facebook.github.io/react-native/releases/0.48/docs/layout-props.html#borderbottomwidth

larsonjj: Ahh, good call. Sorry for the misunderstanding. Should be fixed now

src/jsonUtils/computeYogaNode.js
larsonjj: 👍

jongold:
This is incredible - amazing work @larsonjj. Will try to run it on our real-world stuff this week :)

jacobp100:
So glad this merged! Excellent work @larsonjj 🎉🎉🎉

jongold:
Yeah - amazing work @larsonjj!

I published in 1.0.0-beta.3 — will publish a proper version once everything else is in :)

PR #193, opened 2017-10-10
Add MIT license

larsonjj:
Add MIT license file to make licensing obvious to consumers of this project. Github will also add a nice little badge for easy understanding:

screen shot 2017-10-10 at 3 59 17 pm


jongold:
Doublechecking with @ljharb before merging this — we say MIT in the package.json so I think we're good…

Issue #195, opened 2017-10-17
render with container other than context.document.currentPage fails


weaintplastic:
Hi folks,

I hope it is okay to report issues with the 1.* beta version here.

When using the render function with a container other than `context.document.currentPage()` nothing is rendered.

**Expected behavior:**

I expect the component to be rendered into the given container.

**Observed behavior:**

It renders nothing. This is due to the fact the `render` clears out the document and therefore anything that has been added before (https://github.com/airbnb/react-sketchapp/blob/master/src/render.js#L124). So the actual container is not available anymore when calling `buildPages()`(https://github.com/airbnb/react-sketchapp/blob/master/src/render.js#L130) which would take care about rendering to the given container.

Maybe clearing the document should be optional or the concern of the application that is utilising react-sketchapp.

**How to reproduce:**

```
const onRun = () => {
  const document = sketch.selectedDocument;
  const page = document.selectedPage;
  const group = page.newGroup();

  render(
    <Text>I should be rendered to the group</Text>,
    group
  );

  group.adjustToFit();
};

module.exports = onRun;
```
**Sketch version:**

1.0.0-beta.5

mathieudutour:
> Maybe clearing the document should be optional or the concern of the application that is utilising react-sketchapp.

That would go against the principle of React but we should definitely only clear the object passed as an argument.

There are actually a few places where we hardcoded `context.document` and that's not great. Not sure what the solution for symbols and shared styles tho

weaintplastic:
> only clear the object passed as an argument.

That makes 100% sense. Let me know if I can help out providing functionality for that. probably a function called `resetContainer` rather than `resetDocument`

PR #196, opened 2017-10-19
Reset given container instead of entire document

weaintplastic:
Will fix https://github.com/airbnb/react-sketchapp/issues/195

Resetting the entire document assumes that the container passed to `render` is the current selected page. Whenever a group is passed as the container, render doesn't return anything since the given group will be wiped out when resetting the document.

In order to narrow down the scope of what needs to be reset I've added a function that resets the given container. 

I've left the implementation for resetting the document since I don't quite get the idea behind it. I think keeping the container immutable by resetting and rerender works.

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -6,7 +6,7 @@ import {
 } from 'sketchapp-json-plugin';
 import buildTree from './buildTree';
 import flexToSketchJSON from './flexToSketchJSON';
-import { resetDocument, resetPage } from './resets';
+import { resetDocument, resetContainer, resetPage } from './resets';
 import { getSymbolsPage } from './symbol';
 
 import type { SketchLayer, TreeNode } from './types';
@@ -120,8 +120,8 @@ export const render = (
 ): ?SketchLayer | Array<?SketchLayer> => {
   if (appVersionSupported()) {
     try {
-      // Clear out document to prepare for re-render
-      resetDocument();
+      // Clear out container to prepare for re-render
+      resetContainer(container);
 
       // Build out sketch compatible tree representation
       const tree = buildTree(element);
--- a/src/render.js
+++ b/src/render.js
@@ -1,12 +1,9 @@
 import React from 'react';
 import type { SJLayer } from 'sketchapp-json-flow-types';
-import {
-  appVersionSupported,
-  fromSJSONDictionary,
-} from 'sketchapp-json-plugin';
+import { appVersionSupported, fromSJSONDictionary } from 'sketchapp-json-plugin';
 import buildTree from './buildTree';
 import flexToSketchJSON from './flexToSketchJSON';
-import { resetDocument, resetContainer, resetPage } from './resets';
+import { resetContainer, resetPage } from './resets';
 import { getSymbolsPage } from './symbol';
 
 import type { SketchLayer, TreeNode } from './types';
@@ -31,10 +28,7 @@ export const renderLayers = (layers, container: SketchLayer): SketchLayer => {
   return container;
 };
 
-const renderToSketch = (
-  node: TreeNode,
-  container: SketchLayer
-): SketchLayer => {
+const renderToSketch = (node: TreeNode, container: SketchLayer): SketchLayer => {
   const json = flexToSketchJSON(node);
   const layer = fromSJSONDictionary(json);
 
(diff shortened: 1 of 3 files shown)


mathieudutour:
I think we need a switch if the container is a document or a page/layer. In case of a document, we need to remove the pages (if more than 1) and clear the last one.

@larsonjj I'm actually curious if, when using `Page`, passing the document works? It feels more semantic to me

weaintplastic:
@mathieudutour the api documentation points out that render accepts either a page or a group as container. Is passing document even a use-case, if so I'm happy to add the switch.

mathieudutour:
Yeah I realised that I was writing it. Just wondering if it would make sense to support it: I do see a use case when using `Page`, eg it is weird to pass a page to render into when you define Pages yourself. But I don't think it's supported right now

weaintplastic:
@mathieudutour @larsonjj I'm on vacation without access to the internet for the next two weeks. Let me know if there's anything I can provide beforehand.

larsonjj:
So just to clarify, the current issue is that we are always clearing the entire document on re-render and that doesn’t allow for use cases that want/need multiple mount points within a Sketch document. Let me know if I’m off-base here.   

To me the solution to the issue at hand is close to what you have here @weaintplastic in this PR, but with some tweaks:

 First:

Remove the `resetContainer` function as it’s identical to the `resetPage` function. Then, find/replace the`resetPage` function and rename it to `resetLayer` anywhere it is used.

Second:

We need to conditionalize our clearing logic within the `render` function. If no container is passed into the `render` function, then run the `resetDocument` function. Otherwise run the `resetLayer(container)` function.

Example:

```
// Clear out document or layer to prepare for re-render
if (!container) {
  resetDocument();
} else {
  resetLayer(container)
}

```

With those changes, we should be in a good place.

weaintplastic:
@larsonjj sounds like a plan. There is a tiny difference in `resetContainer` and `resetPage` though. When calling `container.children()` the container itself is somehow part of the result as being the last item in the array. This is why I'm skipping it from being removed. 

https://github.com/airbnb/react-sketchapp/pull/196/files#diff-06e49811ecfd4c1c44b8f6d180d2acaaR36

I couldn't find a reason why it is doing that, maybe you have a clue?

weaintplastic:
@larsonjj @mathieudutour sorry that it took so long to continue working on this. So I've added what you've suggested @larsonjj. Renamed `resetPage` to `resetLayer` and refactored its usage.

I still skip the last child since this is always the container itself when calling `container.children()`. This is true when the container is a `Page` or a `Group`. The only difference is that calling `removeFromParent()` won't have any effect on a `Page` but removes the `Group`. 

Please let me know what you think of this.

weaintplastic:
@larsonjj would you mind taking a quick look?

larsonjj:
Hey @weaintplastic,

Apologies for the delay, was on holiday. Code looks good from an initial glance, will do some testing tonight and verify so we can hopefully get this merged.

weaintplastic:
@larsonjj no worries. hope you had a great vacation. thanks a lot for taking a look.

weaintplastic:
@larsonjj yeah, not really intuitive. But at least the result of `children()` is consistent for pages as well as for groups.

Thanks a lot for your support.

weaintplastic:
@jongold any plans to release a beta version 6 soon which will include this ☝🏻

LincMitch:
Can you provide an example of how to use this in the API docs?

weaintplastic:
@LincMitch this is meant to fix a bug rather than introducing a new feature. The difference is that before this fix react-sketchapp was clearing out the entire document when rendering to a container, regardless of if that container was the page or a group within the page or an artboard. The usecase could be as simple as rendering a compnent in a group rather than the current page as most examples point out:

```js 
const document = sketch.selectedDocument;
const page = document.selectedPage;
const group = page.newGroup();
render(<SomeComponent />, group.sketchObject);
group.adjustToFit();
```

You think providing such an example makes sense? If so I'm happy do add one.

LincMitch:
@weaintplastic Thanks for the code example. 
I think any examples would help, in particular for designers like me. A frequent criticism I have heard from designers is that any update to this plugin via sketch will be lost when the plugin gets updated. Also, React-SketchApp has the potential to attract more designers, and clear examples will help with the steep  Learning curve for designers wishing to code.

Issue #177, opened 2017-09-13
Create `getComponentFromSymbol`


ryngonzalez:
## Proposal

Add the ability to generate a React component that takes in an existing symbol in the Sketch document and returns a component reference for use in other components, either to be rendered directly or in use as an override. Mainly useful since there's no way to render SVGs without primitives (see #16), and this gives us a way to work around it.

### Rendered Directly

```jsx
import PropTypes from 'prop-types'
import React from 'react'
…
const ErrorIconSymbol = getComponentFromSymbol('Icons/ErrorIcon')
const ErrorIcon = ({ size }) => (
  <ErrorIconSymbol
    style={{ width: 24, height: 24 }}
  />
)

const ERROR_ICON_SIZE = 20

const ErrorMessage = ({ error }) => (
  <View>
    <ErrorIcon size={ERROR_ICON_SIZE} />
    <Error>{error}</Error>
  </View>
)

export default ErrorMessage
```

### Used as an override

```jsx
import PropTypes from 'prop-types'
import React from 'react'
…
const ErrorIconSymbol = getComponentFromSymbol('Icons/ErrorIcon')
const WarningIconSymbol = getComponentFromSymbol('Icons/WarningIcon')
const SearchIconSymbol = getComponentFromSymbol('Icons/SearchIcon')
const CalendarIconSymbol = getComponentFromSymbol('Icons/CalendarIcon')

const IconContainer = () => (
  <View
    name="IconContainer"
    style={{ display: 'flex', flexDirection: 'row', justifyContent: 'space-between' }}
  >
    <ErrorIconSymbol
      name="Left Icon"
      style={{ width: 24, height: 24 }}
    />
    <WarningIconSymbol
      name="Right Icon"
      style={{ width: 24, height: 24 }}
    />
  </View>
)

const IconContainerSymbol = makeSymbol(IconContainer);

const Document = () => (
  <Artboard style={{ display: 'flex' }}>
    <IconContainerSymbol
      overrides={{
        'Left Icon': CalendarIconSymbol,
        'Right Icon': SearchIconSymbol,
      }}
    />
  </Artboard>
);

export default (context) => {
  render(<Document />, context.document.currentPage());
}
```

ianhook:
One caveat is, while this project uses name to keep symbols unique, Sketch itself uses a unique ID field. So when searching by name you might not get the symbol you where expecting.

jongold:
👍 👍 👍 👍 👍 👍

PR #198, opened 2017-10-27
Fix symbols

ianhook:
Turns out that #177 was already done, so I updated the docs to reflect that.

There were also 2 problems with overrides that were addressed, so the example now works as expected on v46.

This may also address #112


ianhook:
I've updated the flow types here: https://github.com/darknoon/sketchapp-json-flow-types/pull/10

kmartinezmedia:
Any updates on when this will be merged?

ianhook:
I'm not sure what to do about the flow type requirement which is the major hold up here. Any thoughts @jongold ?

mattborn:
Bump.

ianhook:
So the change to sketchapp-json-flow-types has made it to master, but it hasn't been publish. Once it has, this should be ok to merge.

mathieudutour:
merged as part of #292

PR #207, opened 2017-11-21
fixed main field

smelukov:


jongold:
Yeah this should always point to `lib/index.js` :)

PR #234, opened 2017-12-29
Add SVG link in API.md

thierryc:


mathieudutour:
Good catch!

thierryc:
Thank you. happy new year.

Issue #230, opened 2017-12-21
Performance issue with makeSymbol()


alampros:
I hesitated to submit an issue for this because this project is so amazingly cool and it blows my mind that this is even possible...*however*...

I have a project wherein I am rendering ~50 relatively simple SVGs (~300 symbols) into symbols and it is taking around **20 minutes** to render (all while locking up the sketch process).

Is there something I'm doing here that stands out as super non-performant? Is this just an inherently intensive process that actually *should* take this long?

**Expected behavior:**  
Not instant, but closer to the 2-3 minute range.

**Observed behavior:**  
Render takes between 15-25 minutes.

**How to reproduce:**  

See the demo repo here: https://github.com/alampros/react-sketchapp-svg-render-test

I am importing my SVGs via a [babel plugin](https://github.com/alampros/babel-plugin-react-sketchapp-svg) and calling `makeSymbol` on immediately in the import file like so:

> Note: None of the SVGs are actually being rendered to a page – they are output only to the symbols page (for use as a shared library).

#### `svgImports.js`:
```js
import IconAngleDown from '../svg-tmp/angle-down.svg'
import IconAngleLeft from '../svg-tmp/angle-left.svg'
//... 47 more SVGs imported

export default {
  // Generate 5 color variants for each SVG
  'Black': {
    'angle-down': makeSymbol(() => <IconAngleDown fill="rgb(0, 0, 0)" />, 'Icon / angle-down / Black'),
    'angle-left': makeSymbol(() => <IconAngleLeft fill="rgb(0, 0, 0)" />, 'Icon / angle-left / Black'),
    ...
  },
  'White': {
    'angle-down': makeSymbol(() => <IconAngleDown fill="rgb(255, 255, 255)" />, 'Icon / angle-down / White'),
    'angle-left': makeSymbol(() => <IconAngleLeft fill="rgb(255, 255, 255)" />, 'Icon / angle-left / White'),
    ...
  }
  ...
}
```

#### `main.js`:
```jsx
import { Document, Page, Text, render } from 'react-sketchapp'
import './svgImports.js'

const IconsPage = () => {
  return (
    <Document>
      <Page name="Page 1">
        <Text>See the symbols page</Text>
      </Page>
    </Document>
  )
}

const onRun = () => {
  render(<IconsPage />)
}

module.exports = onRun
```

**Sketch version:** 48.2

[...]

PR #235, opened 2017-12-31
Remove n^2 rendering problem with large symbol sets

ianhook:
The original symbol code required `injectSymbols` to be called separately from `makeSymbols` because `injectSymbols` deletes the Symbols page and recreates it on every call. This can take a while as seen in #230. 

I put the `injectSymbols` call in the render method, because this is the only time it needs to be called.


alampros:
Thanks, @ianhook – this is SOO much faster!

I've added a branch to test this on the demo repo that I created for #230: https://github.com/alampros/react-sketchapp-svg-render-test/tree/test-pull-235

examples/symbols/src/my-command.js
```diff
@@ -6,54 +6,51 @@ import {
   View,
   Image,
   makeSymbol,
+  injectSymbols,
```
alampros: Can this be removed safely? It's throwing an eslint error.

alampros:
This error is thrown on render if `makeSymbol` is never called:

```
TypeError: null is not an object (evaluating 'Object.keys(mastersNameRegistry)')
```

It looks like adding this check (from `makeSymbol`) to `injectSymbols` fixes it:
```js
  if (mastersNameRegistry === null) {
    getExistingSymbols();
  }
```

https://github.com/ianhook/react-sketchapp/blob/2a143bdf0a7c6649e35979d5c5fce976dc72a537/src/symbol.js#L75-L93

src/symbol.js
```diff
@@ -72,36 +72,39 @@ export const getSymbolId = (masterName: string): string => {
...
+    renderLayers(Object.keys(layers).map(k => layers[k]), symbolsPage);
 
-  let notSymbolsPage = array.find(p => String(p.name()) !== 'Symbols');
-  if (!notSymbolsPage) {
-    notSymbolsPage = globalContext.document.addBlankPage();
+    let notSymbolsPage = array.find(p => String(p.name()) !== 'Symbols');
+    if (!notSymbolsPage) {
+      notSymbolsPage = globalContext.document.addBlankPage();
+    }
+    globalContext.document.setCurrentPage(notSymbolsPage);
```
mathieudutour: shouldn't we store the current page at the beginning of the `injectSymbols` method and go back to it instead getting the first one here?

alampros:
What's involved in getting this merged and released?

mathieudutour:
merged as part of #254

PR #237, opened 2018-01-05
Make MSDocument instance's access more compatible

MrPeak:
Error occurred at run time when a sketch action was triggered.

Each `action` has own handler, but not each handler's parameter -- 'context' can access to `document` directly,  for example,  action `documentOpen`.

** Error Message: **

`null is not an object ---> globalContext.document.pages`


ianhook:
The code seems fine, but I don't fully understand the reason for it. For example, why didn't you also make this change [here](https://github.com/airbnb/react-sketchapp/blob/inject-symbols-on-demand/src/render.js#L76)?

MrPeak:
Oh! I truly missed [it](https://github.com/airbnb/react-sketchapp/blob/5485aef6d8e60869ec873fac72f93876675fe176/src/render.js#L75).  That's because I add the Layer which is created by `sketchapp-json-plugin` to another Layer (or Page) manually, not via 'render' method.

Under some circumstances，`context` doesn't include `document` object, that's why I make this compatible operation.

MrPeak:
So, what's going on now?

PR #244, opened 2018-01-16
fix(reactTreeToFlexTree): Fix layout for absolute layers with zIndex

weaintplastic:
This PR is fixing a bug where the yoga layout is applied in the wrong order for layers that are positioned absolute.

- Added a fix
- In order to test the fixed bahavior reactTreeToFlexTree is now exported.

### DIFF
--- a/src/buildTree.js
+++ b/src/buildTree.js
@@ -9,7 +9,7 @@ import computeTextTree from './jsonUtils/computeTextTree';
 import { INHERITABLE_FONT_STYLES } from './utils/constants';
 import zIndex from './utils/zIndex';
 
-const reactTreeToFlexTree = (
+export const reactTreeToFlexTree = (
   node: TreeNode,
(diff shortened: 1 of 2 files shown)


src/buildTree.js
```diff
@@ -52,8 +52,8 @@ const reactTreeToFlexTree = (
     for (let index = 0; index < children.length; index += 1) {
       const childComponent = children[index];
       const childStyles =
-        childComponent.props && childComponent.props.styles
-          ? childComponent.props.styles
+        childComponent.props && childComponent.props.style
```
weaintplastic: Due to the typo the `childStyles` always have been an empty object rather than the actual style. As a result the `decrementIndex` condition couldn't detect absolute positioned layers.

weaintplastic:
@larsonjj this is now ready to be merged and published! Thanks so much for the approval!

PR #246, opened 2018-01-18
Create page only when currently on symbols page

dchen:
Previously, rendering would automatically create a new page if a symbols page exists at all, regardless of whether it is currently selected. This changes the behavior to use the current page when it is not the symbols page.

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -73,7 +73,8 @@ const buildPages = (
 
   if (pageData.length === 0) {
     const _container = container || context.document.currentPage();
-    const page = !symbolPage ? _container : context.document.addBlankPage();
+    const page =
+      _container !== symbolPage ? _container : context.document.addBlankPage();
 
     return renderToSketch(tree, page);
   }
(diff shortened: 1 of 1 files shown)


larsonjj:
@dchen, the logic in this PR makes sense. Just wondering what the use case for this is? Does it ever make sense to pass in the Symbols page as the container to `render`?

dchen:
@larsonjj This restores the ability to render something simple to the current page on a document with symbols. 

When rendering something without pages, the previous version would render to the current page. The current code renders to the current page only if a symbols page does not exist, otherwise it creates a new page. I think that it makes more sense instead to render to the current page unless the symbols page is the current page (in this case, we create a new page).

For example, currently if you run the basic-setup example on a document with a symbols page with Page 1 selected, it renders to a new page. If you run the same example on a document without a symbols page, it renders to the current page. With this PR, it would always render to the current page unless you're on the symbols page.

larsonjj:
Ahh, I see. Thanks for the explanation!

PR #178, opened 2017-09-15
Multiple Pages and Artboards Support

larsonjj:
# Overview

This PR adds the ability to create additional sketch pages and artboards. Will also keep backwards compatibility with existing code.

# Example
```jsx
render(
  <Document>
    <Page name="Mobile">
      <Artboard name="iPhone 5 SE">
        <View>
          <Text>Hello World</Text>
        </View>
      </Artboard>
      <Artboard name="iPhone 6S">
        <View>
          <Text>Hello World Again</Text>
        </View>
      </Artboard>
    </Page>
    <Page name="Tablet">
      <Artboard name="iPad">
        <View>
          <Text>Hello Second World</Text>
        </View>
      </Artboard>
    </Page>
  </Document>
);
```

# Changes

- Added `Document` and `Page` components
    - `Document` components are really just dummy placeholders to allow for multiple `Page` component children
    - Pages can have custom names, but will avoid `Symbols` and rewrite it to `Symbols (renamed to avoid conflict` so as to not overwrite symbols pages
    - New Jest tests and snapshots were added to cover these new components
    - Updated API docs to reflect usage
    - Added some more useful error messages on render failures to help users understand how to fix them in most cases

# Related Discussions

- https://github.com/airbnb/react-sketchapp/issues/139
- https://github.com/airbnb/react-sketchapp/issues/17

PR #256, opened 2018-02-08
Fix #255 page not supporting properties

### LINKED ISSUE #255: Page component doesn't support style prop
_I am..._

| Reporting a bug or issue

**Expected behavior:**

```Page``` component should accept ```style``` prop and pass to underlying ```page``` component.   Maybe even just pass all props.

Without the ability to pass ```style```, Artboard position on the canvas cannot be controlled.

**Observed behavior:**

Currently only renders ```Page``` children into the ```page``` node.  Any props passed to ```Page``` are not passed to ```page```

**How to reproduce:**

```
render(<Document>
  <Page style={{ backgroundColor: 'red' }}>
    <Artboard>
      <View>
        <Text>Hello World</Text>
      </View>
    </Artboard>
  </Page>
</Document>)
```

NB: ```render``` is not supplied a context so that it renders all pages in the JSX

**Sketch version:**

48.2

### PR DESCRIPTION
**Pull request by dmeehan1968**
Fix #255 

Pass properties to ```Page``` component through to sketch ```page```, essentially to handle passing of style properties which can be useful in positioning Artboards within a Page.

```
<Document>
  <Page style={{ flex: 1, flexDirection: 'row' }}>
    <Artboard></Artboard> 
    <Artboard></Artboard>
  </Page>
</Document>

```

`style` property is flattened before passing, other properties are passed 'as is'.

### CODE REVIEW

File: src/components/Page.js [hunk 1]
```diff
@@ -13,11 +25,18 @@ class Page extends React.Component {
...
   render() {
-    const { name, children } = this.props;
-    const _name =
-      name === 'Symbols' ? 'Symbols (renamed to avoid conflict)' : name;
+    const { name, children, style, ...otherProps } = this.props;
 
-    return <page name={_name}>{children}</page>;
+    const pageProps = {
+      name: name === 'Symbols' ? 'Symbols (renamed to avoid conflict)' : name,
+      ...otherProps,
```
mathieudutour: can we avoid using `...` twice? just pass the `name` prop without merging it with the `otherProps`. Same for `style` I'd say

dmeehan1968: I can do, but was trying to avoid changing the snapshots for the existing tests.

The following removes the additional spread operator, but causes the existing tests to fail because ```style``` becomes an property with undefined value.  I don't know how the underlying Sketch ```page``` component handles that and it could have an unintended side effect.

Making ```style``` conditionally included in the ```pageProps``` var avoids this possibility.

Let me know what you think, I'll update the PR if you want.

```
  render() {
    const { name, children, style, ...otherProps } = this.props;
    const _name = name === 'Symbols' ? 'Symbols (renamed to avoid conflict)' : name;
    const _style = StyleSheet.flatten(style);

    return <page name={_name} style={_style} {...otherProps}>{children}</page>;
  }
```

mathieudutour: yeah I think it's fine, just update the existing snapshots.

Have you tested manually that it worked? (I usually use the `basic-setup` example and use `../../lib/` instead of `react-sketchapp` so I can test my changes directly)

dmeehan1968: Yep, tested on Basic setup, although that doesn't specifically allow me to test the changes.  I'm happy the unit test validates the intention.

LincMitch:
Great - so we now have horizontal Artboards. This issue was raised in #178.

dmeehan1968:
There seemed to be a problem with artboards growing to equal width on the secondary axis, which I would have thought ```flexShrink: 1``` would solve, but didn’t. I didn’t play around with it for long though. 

Sketch seems to default to ```flex: 1``` on all components. 

It would be interesting to know if we could access Sketch’s artboard presets programatically. E.g.

```
<Artboard preset=”iPhone X”></Artboard>
```

SUMMARY.md
```diff
@@ -1 +0,0 @@
-docs/README.md
```
mathieudutour: what is this change?

dmeehan1968: I merged master to make sure your merge would be easier, as it said it was out of date.  It looks like SUMMARY.md didn't have a newline at the end, that's what git diff seems to be reporting, but as a result it seems to think that all the content is new.  I didn't touch the file, so not sure why that's happened.

mathieudutour: hum weird, it looks like it followed the symlink 🤔
Could you fix it? I'll merge once it's done

dmeehan1968: I'm not entirely sure I know how.  Any suggestions?

mathieudutour: I'd say `git checkout master -- SUMMARY.md` or something similar

dmeehan1968: Done

PR #264, opened 2018-02-15
(docs) Add detailed "slow-start" for new folks

### LINKED ISSUE #263: (docs) Add a "slowstart"
| Requesting a new feature

**Proposal/Feature-request:** Detailed starter docs for non-technical/inexperienced folks to get up and running.

I was having dinner with a non-technical designer person and they mentioned they were having quite a bit of trouble getting the basic examples to work -- the "quickstart" commands in particular were intimidating/confusing. There was also confusion about when to open Sketch and how all the commands/pieces fit together. They got snared/blocked on some pretty uninteresting issues.

Rather than pollute the root README with a bunch of instructions, a better approach seems to be to enhance the docs/examples.md with a bit more hand holding so a confused user can get the concept and reach a good jumping off point.

If approved, there'd be both a quickstart for those who get it + a slow start for those who need it

valgaze:
If approved, resolves #263


docs/examples.md
```diff
@@ -32,3 +34,109 @@
...
+### Step 1: Clone react-sketchapp
+
+Assuming all your tools are installed/working, we're ready to get started by getting ```react-sketchapp``` onto your machine
+
+- [ ] Open a terminal and change the directory into a folder where you want react-sketchapp to live (if you're not sure or don't really care, an easy choice is to put it right on your desktop. You can move your terminal inside there by typing ```cd ~/Desktop``` & verify you're inside that directory by typing ```pwd``` to see the full path. ```cd ..``` will move you back one folder)
+
+- [ ] You now can clone or download react-sketchapp to the directory from the step above by copying the following command into your terminal:
+
+```sh
+git clone https://github.com/airbnb/react-sketchapp.git
```
mathieudutour: it's a bit going against the instructions given in each examples: curl each example separetely so the people are not scared by all the other files

mathieudutour:
Thanks for the docs! I'm not sure about the "clone everything" approach tho

docs/examples.md
valgaze: I suppose it does break the convention slightly, however, my sense is that just to get them up and running it's a much cleaner approach to pull in the entire repo. Each of the individual examples have a line that say "Download the example or clone the repo" and the distinction might be lost on new folks. If they run through the instructions they'll probably pick up how to navigate with the command line and can jump off from there.

 Then individual clone commands with pipes to things like ``` | tar -xz --strip=2 ``` or ```curl``` would make more sense

valgaze: @mathieudutour What are you thinking?

Issue #161, opened 2017-08-03
Feature to support multiple shadows


weaintplastic:
Hi 👋🏻 ,

I'm Roland and recently fell in love with the great work that you're doing here and the endless possibilities that come with it. While already having some PR's running (you've probably noticed), I'd like to request a feature or better discuss the approach to introduce it.

Currently it is only possible to define a single box shadow. I'd like to add a feature to support multiple ones. While sketch already provides setting multiple shadows by passing in shadow styles in an array, the ViewRenderer as well as ImageRenderer do only support resolving the style for one shadow.

https://github.com/airbnb/react-sketchapp/blob/master/src/renderers/ViewRenderer.js#L48
https://github.com/airbnb/react-sketchapp/blob/master/src/renderers/ViewRenderer.js#L104

As a proposal (happy to discuss) I'd suggest to create a style property called `shadowGroup` in parallel. Each of them holding styles for an individual shadow. I'd still keep the current implementation in place for backwards-compatibility. So creating multiple shadows would be an opt-in functionality. 

Let me know what you think. 

Roland

ianhook:
While this probably will work great in Sketch, it will not work correctly when rendered anywhere else. Setting aside whether this should allow users access to the full power of sketch, if it does do, it  should do so in a way that users unfamiliar to React Native can use the component with full knowledge of what they're getting themselves into.

jongold:
Thanks for the proposal Roland! My thoughts === Ian's — we're designing the
API of react-sketchapp to mirror react-native (or actually
react-primitives). That said, there are obviously Sketch-specific features
that we want to support too - Artboards & Symbols are other examples.

I appreciate you keeping backwards compatibility — I think it's low enough
risk that we can include it (with the caveat that it won't work
cross-platform if you're using react-primitives), and iterate on the API
later if needed.

Should we leave this open for a little longer before merging to garner more
feedback?

jon.gold
628.228.1152

weaintplastic:
Thank's so much providing some background on your opinions. Really appreciate that. I totally agree to leave this open for a little while longer and see what others think. 

Roland

mrtnbroder:
I'd agree that this project here should add as much sketch features as possible and not care so much about the cross-platform integrations. I mean, it is a fine goal to achieve cross-platform support, but since sketch is a totally different program, it is just not possible. I'd like to get all the power of sketch within this project.

LFra:
[...]

PR #277, opened 2018-03-03
Add possibility to add multiple shadows and shadow spread

ludwigfrank:
As discussed in #161 this would add the shadowGroup prop to View, allowing multiple shadows to be added.

P.S. I'm new to working on big projects like this. I would be very grateful for any feedback. Thanks for opening this project to the community.


src/components/View.js
```diff
@@ -18,6 +19,11 @@ const propTypes = {
   resizingConstraint: PropTypes.shape({
     ...ResizingConstraintPropTypes,
   }),
+  shadowGroup: PropTypes.arrayOf(
```
mathieudutour: let's just call it `shadows`?

src/components/ShadowGroupPropTypes.js
```diff
@@ -0,0 +1,9 @@
+import PropTypes from 'prop-types';
+
+module.exports = {
+  shadowColor: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
mathieudutour: we should add `shadowInner` to support multiple inner shadows as well

src/jsonUtils/shapeLayers.js
```diff
@@ -2,7 +2,7 @@
 import type { SJFill, SJPath, SJRect, SJShapeGroupLayer } from '@skpm/sketchapp-json-flow-types';
 import { makeResizeConstraint } from './hacksForJSONImpl';
 import { generateID, makeRect } from './models';
-import type { ResizeConstraints } from '../types';
+import type { ResizeConstraints, SketchShadowGroup } from '../types';
```
mathieudutour: used anywhere?

src/renderers/ViewRenderer.js
```diff
@@ -106,7 +113,13 @@ class ViewRenderer extends SketchRenderer {
     const fill = makeColorFill(backgroundColor);
     const content = makeShapeGroup(frame, [shapeLayer], [fill]);
 
-    if (hasAnyDefined(style, SHADOW_STYLES)) {
+    if (props.shadowGroup) {
+      const shadows = [];
+      props.shadowGroup.map(shadowStyle =>
+        shadows.push(makeShadow(shadowStyle))
```
mathieudutour: let's push to `innerShadows` as well if `shadowStyle.shadowInner` is `true`

src/renderers/ViewRenderer.js
```diff
@@ -106,7 +113,13 @@ class ViewRenderer extends SketchRenderer {
...
+      );
+      content.style.shadows = shadows;
+    } else if (hasAnyDefined(style, SHADOW_STYLES)) {
```
mathieudutour: instead of an `else if` we should probably concatenate the shadow defined in `style` if there is one

src/types.js
```diff
@@ -73,6 +73,8 @@ export type ViewStyle = {
   shadowOffset: { width: number, height: number },
   shadowOpacity: number,
   shadowRadius: number,
+  shadowSpread: number,
+  shadowInner: "innerShadow" | "shadow",
```
mathieudutour: this should be a `boolean`.

We could probably reuse the `SketchShadow` type as well (`{...} & SketchShadow`)

thierryc:
#358 fix conflicts.

Thank you @ludwigfrank

mathieudutour:
fixed in #358

PR #291, opened 2018-04-17
Add Canvas component (enables rendering multiple Artboard's on existing page)

jaridmargolin:
## Problem

There are currently two options for rendering using react-sketchapp:

1. Render the entire document using the `<Document>`, `<Page>`, and `<Artboard>` components
2. Render to an existing page (defaults to current page, but can also be explicitly passed in)

Unfortunately, rendering to an existing page has a few restrictions as pointed out in the docs:

> Artboard: Wrapper for Sketch's Artboards. Requires a `<Page>` component as a parent if you would like to use multiple of these components.

I'm not entirely sure if this restriction was put in place for a reason, but I found it to be limiting. My use case is fairly simple, I would like to render multiple `Artboard`s on a single page, without having to control the entire document.

## Solution

I would like to purpose the introduction of a new component, `Canvas`, to accomplish this. While there is no equivalent component referenced in the [Sketch API](https://developer.sketchapp.com/reference/api/), the concept is introduced and described in the [Sketch Docs](https://www.sketchapp.com/docs/the-interface/canvas/).

**Example Usage**:

```
class CustomPage extends Component {
  render () {
    return <Canvas>
      <Artboard name='Artboard 1'>
        <Text>Text1</Text>
      </Artboard>
      <Artboard name='Artboard 2'>
        <Text>Text2</Text>
      </Artboard>
    </Canvas>
  }
}

export default (context) => {
  const doc = sketch.fromNative(context.document)
  const customPage = doc.pages.find(page => page.name === 'Custom')

  render(<CustomPage />, customPage.sketchObject)
}
```

### Implementation Notes

 I chose to add this functionality inside of the `renderToSketch` method so that the `Canvas` element can be used in both contexts, when rendering to an existing page, and when rendering the entire document. I felt like allowing it to be used in both contexts would result in the least confusion. It simply introduces itself into the document Hierarchy (Document, Page, Canvas, Artboard/SymbolMaster).

I also explored the idea of adding support for multiple `Artboard`s by allowing the root element to be a `React.Fragment`. I was able to get this approach to work (and possibly this should be supported no matter what), however the `Canvas` is slightly more convenient approach as it allows the `Artboard`s to be aligned using FlexBox.

---

If this idea is of interest, I would elaborate on this PR by adding documentation, and additional tests (although it doesn't appear there are existing tests for the render code flow which I needed to alter in order to support this).

### DIFF
--- a/src/components/Canvas.js
+++ b/src/components/Canvas.js
@@ -1,0 +1,36 @@
+/* @flow */
+import React from 'react';
+import PropTypes from 'prop-types';
+import StyleSheet from '../stylesheet';
+import CanvasStylePropTypes from './CanvasStylePropTypes';
+
+const propTypes = {
+  children: PropTypes.node,
+  style: PropTypes.oneOfType([
+    PropTypes.shape({ ...CanvasStylePropTypes }),
+    PropTypes.arrayOf(
+      PropTypes.oneOfType([
+        PropTypes.shape({ ...CanvasStylePropTypes }),
+        PropTypes.number,
+      ])
+    ),
+    PropTypes.number,
+  ]),
+};
+
+class Canvas extends React.Component {
+  render() {
+    const { children, style, ...otherProps } = this.props;
+    const _style = StyleSheet.flatten(style);
+
+    return (
+      <canvas style={_style} {...otherProps}>
+        {children}
+      </canvas>
+    );
+  }
+}
+
+Canvas.propTypes = propTypes;
+
+module.exports = Canvas;
--- a/src/components/CanvasStylePropTypes.js
+++ b/src/components/CanvasStylePropTypes.js
@@ -1,0 +1,1 @@
+module.exports = {};
--- a/src/index.js
+++ b/src/index.js
@@ -4,6 +4,7 @@ import Platform from './Platform';
 import StyleSheet from './stylesheet';
 import Document from './components/Document';
 import Page from './components/Page';
+import Canvas from './components/Canvas';
 import Artboard from './components/Artboard';
 import Image from './components/Image';
 import RedBox from './components/RedBox';
@@ -19,6 +20,7 @@ module.exports = {
   StyleSheet,
   Document,
   Page,
+  Canvas,
   Artboard,
   Image,
   RedBox,
--- a/src/render.js
+++ b/src/render.js
@@ -36,6 +36,10 @@ const renderToSketch = (
   node: TreeNode,
   container: SketchLayer
 ): SketchLayer => {
+  if (node.type === 'canvas') {
+    return node.children.map(child => renderToSketch(child, container));
+  }
+
   const json = flexToSketchJSON(node);
   const layer = fromSJSONDictionary(json);
 


mathieudutour:
Do you think it would be possible to use a `React.Fragment` and if it's the root, then position the artboards using flexbox automatically?

If not, I'm wondering if we should force a Canvas between a Page and the Artboards.
And If so, we should remove the style property from the Page component and add a check that a Page have a Canvas.

jaridmargolin:
@mathieudutour

> Do you think it would be possible to use a React.Fragment and if it's the root, then position the artboards using flexbox automatically?

The thing about `Page` and `Canvas` is that they allow you to define styles so you can control things such as flex direction. `React.Fragment` is really just a shorthand for returning an array of elements, so it wouldn't support receiving styles.

> If not, I'm wondering if we should force a Canvas between a Page and the Artboards.
And If so, we should remove the style property from the Page component and add a check that a Page have a Canvas.

Forcing such a restriction would be a breaking change, but perhaps would make the most sense (at the cost of also being slightly more verbose).

jaridmargolin:
I noodled on this idea, but I error towards minimizing the amount of "AHA moments" an engineer has to endure, and I think having the `Page` component work differently depending on the context it is used, may result in some "AHA moments".

This is really just a personal preference. If you are bullish on not introducing `Canvas`, I can explore what it would take to have `Page` be context aware.

mathieudutour:
I mean it’s probably what you would expect anyway, isn’t it? What would inserting a page into a page do otherwise? (Actually curious what it’s doing at the moment)

jaridmargolin:
At the moment it throws with an error: `Error: Could not find renderer for type 'page'`.

I probably wouldn't find myself creating a `Page` within a `Page`, but it could introduce some unexpected behavior if someone did. Say for example you set the `name` on the nested `Page`, would you expect it to change the name of the page in which it was inserted or would it simply get ignored?

mathieudutour:
I’d say it should update it but it’s a detail and I don’t really mind either way :)

jaridmargolin:
> I’d say it should update it but it’s a detail and I don’t really mind either way :)
> I mean it’s probably what you would expect anyway, isn’t it?

So this is what I was trying to describe. Using the `Canvas` component, these edge cases and accompanying assumptions/preferences get completely bypassed. We don't have to worry about providing additional docs to the `Page` component to describe how it behaves in each scenario. We can just introduce a new Component that provides additional functionality.

I could keep trying to illustrate why I prefer the introduction of the `Canvas` component, but for the sake of moving forward, I'd be willing to implement whatever you deem to be the right solution so that I can render multiple Artboards on a single page.

Have I convinced you of the merits of introducing a new `Canvas` component, or do you would you like me to close this PR and open a new PR that instead alters `Page` behavior?

mathieudutour:
Well the thing about `Canvas` is that is doesn't map to any kind of Sketch concept, the only use case for it is when you want to overwrite the current page with a bunch of artboards. I'm saying that if we can fix that use case using the current API, we should do that instead of adding another concept.

PR #292, opened 2018-04-19
Fix symbol inconsistencies.

jaridmargolin:
This is based off of #198 which includes fixes to get Symbols working within the last versions of Sketch.

## Problem

The example listed in the docs for overriding text elements is not working. Upon further investigation this is an issue with text overrides only working when the name is explicitly defined on the`<Text>` element that needs to be overwritten.

## Solution

The error above occurs due to how overrides are extracted from symbols.

```js
    // here we're doing some look-ahead to see if this group contains a group
    // that contains text. this is the structure that will appear if the user
    // creates a `<Text />` element with a custom name
    const subGroup = subLayer.layers.find(l => l._class === 'group');
    const textLayer =
      subGroup && subGroup.layers
        ? subGroup.layers.find(l => l._class === 'text')
        : null;
    if (textLayer) {
      output.push({
        type: 'text',
        objectId: textLayer.do_objectID,
        name: subGroup.name,
      });
      return;
    }
```

Here, the `subGroup.name` is utilized as the override name rather than the `textLayer.name`. As a result, if no name is passed to the `Text` component, the override name is set to the default name of the group, `Text`.

By using the `textLayer.name` instead, which defaults to the text content, we can the use the text content as the key of the override (as shown in the documentation). This also makes more sense when manually reviewing Sketch, where you would expect the override key you define on the symbol instance to match the name of the override that you see in Sketch.

... unfortunately, this resulted in breaking overrides where the name is explicitly set on the `Text` component.

This is due to the `textLayer` name always being set to the text content. If instead, we pass the name defined on the `Text` component all the way down to the `textLayer` (which to me would be the expected behavior), we can once again correctly override named `Text` components.

### Solution Notes

Looking at additional components (Photo, View, etc..), none of them pass the name down to the rendered layer. I'm not entirely sure what the reasoning would be, but it feels intentional.

If in-fact this is intentional, and the name is not intended to be passed down to the rendered layer, an alternative solution, could be to set the default group name of the `Text` component to the text content rather than the string, "Text."


jaridmargolin:
I have pushed an update that removes the default `name` prop from the `Text` component, and all appears to work when manually testing.

Removing the default `name` prop  should have no consequences due to the `TextRenderer` manually setting the default name for the group:

```js
  getDefaultGroupName(props: any) {
    return props.name || 'Text';
  }
```

mathieudutour:
is this ready for review?

mathieudutour:
so I re-added `injectSymbols`. Feel free to open another PR to remove it if you feel like it but I really want to publish now

jaridmargolin:
Anything related to `injectSymbols` was done in the the work of @ianhook to fix symbols. The changes I made simply adjusted the naming conventions when rendering text layers.

Thanks for the merge.

PR #293, opened 2018-04-20
tests: use npm ci

DanielRuf:
`npm ci` is much faster and is meant for CI. This PR is meant for evaluating the new ci command and testig the performance and compare the logs of Travis CI.

https://docs.npmjs.com/cli/ci
http://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,5 +9,4 @@ _book
 
 # only apps should have lockfiles
 yarn.lock
-package-lock.json
 npm-shrinkwrap.json
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,1 @@
-package-lock=false
 yoga-layout:platform=browser
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,10 @@ notifications:
 node_js:
   - '7'
   - '6'
+before_install:
+  - npm i -g npm@latest
+install:
+  - npm ci
 before_script:
   - npm prune
 script:
--- a/circle.yml
+++ b/circle.yml
@@ -2,7 +2,9 @@ general:
   branches:
     ignore:
       - gh-pages
-
+install:
+  - npm i -g npm@latest
+  - npm ci
 machine:
   environment:
     PATH: "${PATH}:${HOME}/${CIRCLE_PROJECT_REPONAME}/node_modules/.bin"
--- a/circle.yml
+++ b/circle.yml
@@ -2,13 +2,12 @@ general:
   branches:
     ignore:
       - gh-pages
-install:
-  - npm i -g npm@latest
-  - npm ci
 machine:
   environment:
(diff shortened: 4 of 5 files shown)


mathieudutour:
Hum from what I understood, we shouldn’t check in the package-lock.json for a library. Also this doesn’t look any faster

ljharb:
both correct; only apps should have lockfiles, and npm ci is intended only for projects with lockfiles.

DanielRuf:
This is not the case anymore since npm 5 as lockfiles should be always committed and it's done in many big(ger) projects.

The ci command is meant for CI (Travis, CircleCI, appveyor, ...) ;-)

Anyway, this PR was meant to evaluate the performance of it.

ljharb:
No, npm 5’s lockfile should absolutely not be committed or even enabled on non-apps, despite their stance on it. Only apps should have lockfiles, forever.

DanielRuf:
Also see https://github.com/sindresorhus/ama/issues/479#issuecomment-311039385 ;-)

ljharb:
I’ve read that article and discussed it with its author personally; i continue to think it’s very wrong. Only apps should have lockfiles.

boriskor:
@ljharb , can you elaborate why do you think so, please? Also, if my library only has devDependencies/peerDependencies, do you also think it's bad to have a lockfile?

ljharb:
@boriskor yes, it’s still bad, because consumers see peerDeps.

First, there’s two kinds of lockfiles: dev-only, like yarn.lock and package-lock.json, and published - npm-shrinkwrap.json (altho yarn ignores it).

The reason why i think it’s a bad idea is because (for deps and peerDeps), if you have a dev-only lockfile, then you’re seeing a dependency tree that’s frozen in time, and doesn’t match what your users see - they’ll get the (potentially broken) up-to-date tree, while you sit in blissful ignorance with your frozen tree.

If you have a published lockfile, then while your npm users at least see the same deps as you do (a plus), you’ve now obstructed your users from being able to maximally deduplicate their dep graph, likely condemning them to larger node_modules, larger browser bundles, and slower install times. Using a published lockfile (in a non-app) is user-hostile, and always a bad idea (note that a stand-alone binary that has no “main” with no ecosystem - which excludes, say, eslint or gulp - is an “app”, and it would be fine, altho atypical, to publish a shrinkwrap with it).

Without a lockfile, you’ve ensured that a) you and your users have the same experience when installing at the same point in time, leading to easier debugging and a higher chance of discovering s bug in the graph, and b) allowing for maximal deduping, which makes disk space and bundle sizes and install times faster.

If a dev-only lockfile can be used only for dev deps (excluding ones that are also in peer deps), then that would be fine too.

DanielRuf:
Hm, and .npmignore would not be a solution for consumers?

ljharb:
npmignore is solely used by `npm publish` to determine what files to filter out of the package when publishing, it’s entirely irrelevant to the dep graph.

PR #296, opened 2018-04-21
Remove `defaultProps` from `Page` component (enables auto-incrementing page name)

jaridmargolin:
## Problem

The default name prop on the `Page` component provides no value. It in-fact has negative consequences. If we create multiple pages with no specified name, they will all default to `Page 1`. If instead we let the name remain undefined, Sketch will create the page using an auto-incremented name (Page 1, Page 2, Page 3, etc...)

## Solution

Remove the `defaultProps.name` definition :)

### Implementation Notes

This is a larger concern than just this particular PR, but I don't see where or how I can create tests to confirm functionality that involves Sketch's behavior (in this case, the  behavior of `document.addBlankPage`). I was able to manually test this by opening sketch and rendering multiple `Page` components without specifying the name.

### DIFF
--- a/__tests__/components/Page.js
+++ b/__tests__/components/Page.js
@@ -30,12 +30,6 @@ describe('<Page />', () => {
       expect(tree).toMatchSnapshot();
     });
 
-    it('defaults to Page', () => {
-      const tree = renderer.create(<Page />).toJSON();
-
-      expect(tree).toMatchSnapshot();
-    });
-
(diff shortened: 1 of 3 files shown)


mathieudutour:
> where or how I can create tests to confirm functionality that involves Sketch's behavior

I don't think there is a way to do it now but if you want to look at it, have a look at https://github.com/skpm/skpm/tree/master/packages/test-runner

PR #297, opened 2018-04-21
Rework render flow.

### LINKED ISSUE #262: Multiple pages are created when using makeSymbol after saving
| Reporting a bug or issue

**Observed behavior:**

When using symbols, while running `npm run render`, every time I save any new changes, a new page is created in sketch. This doesn't happen if makeSymbol is not used, but if makeSymbol was ever used, even if it's removed, it keeps happening unless stopping process and running it again.

**How to reproduce:**

Run any example that uses makeSymbol like the one @jongold created [here](https://github.com/airbnb/react-sketchapp/blob/master/examples/symbols/src/my-command.js), make any change while running live reloading 

**Sketch version:** 

48.2

Empty Project
image

After running `npm run render`, Page 2 is created

image

After saving, Page 3 is created

image

Is this how is supposed to work?

jaridmargolin:
[Edited: 2018/05/07]

## Breaking Changes (kind of)

* Now throws if the "Symbols" page is explicitly passed in as the `container` on the `render` method. Previously if you explicitly passed in the "Symbols" pages as a container, it would create a new page and render onto that. I would consider that to be unexpected behavior.
* Now throws an error if you attempt to render a `Document` component into a node intended to be a child of Document. Previous behavior was unexpected. It would result in reseting the currently selected page only, and then it would additionally render all child pages defined within the passed `Document`. This would essentially result in pages indefinitely being added to the document upon calling render. This may possibly close #262.

## Enhancements

* More predictable rendering of `RedBox`. The previous implementation would not respect the passed in container. It would always render the `RedBox` into the currently selected page. This could have real negative consequences, by accidentally rendering over pages that were manually constructed. The new behavior will first check for a passed in container, if it does not exist, it will either render onto the currently selected page, or it will create a new page if the currently selected page is the "Symbols" page.
* Adds support for rendering a `Page` component into a container passed through the `render` method. This allows for rendering multiple `Artboard`s onto an existing page. This would close #291 as there would no longer be a need for the `Canvas` component.

---

## Additional Notes

* As mentioned in #296: This is a larger concern than just this particular PR, but I don't see where or how I can create tests to confirm functionality that involves Sketch's behavior 
* I altered a few pieces of puzzling code. Because the render code path does not have automated tests, it is hard to confirm if this will result in any regressions. I have left inline comments on Github in places where I was uncertain.

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -4,11 +4,13 @@ import { appVersionSupported, fromSJSONDictionary } from 'sketchapp-json-plugin'
 import buildTree from './buildTree';
 import flexToSketchJSON from './flexToSketchJSON';
 import { resetLayer, resetDocument } from './resets';
-import { getSymbolsPage, injectSymbols } from './symbol';
+import { injectSymbols } from './symbol';
 
-import type { SketchLayer, TreeNode } from './types';
+import type { SketchDocument, SketchLayer, SketchPage, TreeNode } from './types';
 import RedBox from './components/RedBox';
 import { getDocumentFromContainer, getDocumentFromContext } from './utils/getDocument';
+import isNativeDocument from './utils/isNativeDocument';
+import isNativeSymbolsPage from './utils/isNativeSymbolsPage';
 
 export const renderToJSON = (element: React$Element<any>): SJLayer => {
   const tree = buildTree(element);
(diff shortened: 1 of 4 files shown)


src/resets.js
```diff
@@ -1,28 +1,25 @@
 // @flow
+import isSymbolsPage from './utils/isSymbolsPage';
 
 export const resetLayer = (container: Object) => {
-  const layers = container.children();
-  // Skip last child since it is the container itself
-  for (let l = 0; l < layers.count() - 1; l += 1) {
-    const layer = layers.objectAtIndex(l);
-    layer.removeFromParent();
-  }
```
jaridmargolin: This was especially puzzling to me.

From a code standpoint, I don't understand "Skip last child since it is the container itself"... The values of `container.children()` are only the containers children. It would make no sense for `container.children()` to contain itself.

When manually running this piece of code, it did exactly what I expected it to... It left a child on the page. To be more clear. If I create a page, I insert and Artboard on that page, and I run `restLayer` on that page, the Artboard is not removed.

src/render.js
```diff
@@ -42,104 +42,74 @@ const renderToSketch = (
...
+    throw Error('Cannot render into Symbols page');
+  }
+
+  const doc = getDocument(context);
+  const currentPage = doc.currentPage();
+  try {
+    const tree = buildTree(element);
+
+    if (tree.type === 'document' && container) {
+      throw Error('Cannot render `Document` component into a container');
```
mathieudutour: I don't think that's right. What if I pass a document as the container? I might not want to render into the current document

src/render.js
```diff
@@ -42,104 +42,74 @@ const renderToSketch = (
...
 
-    if (data.children && data.children.length > 0) {
-      // Clear out page layers to prepare for re-render
-      resetLayer(page);
-      data.children.forEach((child) => {
-        renderToSketch(child, page);
-      });
-    }
+  return tree.children.map((child, i) => {
+    const page =
```
mathieudutour: shall we add a check to make sure the child is a Page?

src/render.js
```diff
@@ -42,104 +42,74 @@ const renderToSketch = (
...
-    const _container = container || getDocument(context).currentPage();
-    const page =
-      !symbolPage || _container !== symbolPage
-        ? _container
-        : getDocument(context).addBlankPage();
+  if (tree.type === 'page') {
+    // assume if name is set on this nested page, the intent is to overwrite
+    // the name of the page it is getting rendered into
+    if (tree.props.name) {
+      container.setName(tree.props.name);
```
mathieudutour: :+1:

src/render.js
jaridmargolin: We can... would you want to throw if it isn't?

mathieudutour: Yeah that'd be nice

src/resets.js
jaridmargolin: Must be a user error of some sort, but upon further investigation this check does appear necessary when using `layer.children()`. However, we may be able to bypass this check by instead utilizing `layer.layers()` which appears to return only the direct descendants of the layer.

jaridmargolin:
Yep. After reviewing #254 it is clear that this work should be rebased ontop of it. I will wait for it's merge and then push appropriate updates. Thanks for the set of 👀@mathieudutour

jaridmargolin:
@mathieudutour - I have refreshed this PR of off current master (which now includes #254). When you have a moment, could you please give this a look?

PR #301, opened 2018-04-23
add support for Sketch 50 and fix text style matching

thierryc:
Mathieudutour patch 1 1 #290


thierryc:
do cleaner request.

thierryc:
do cleaner request

PR #310, opened 2018-05-02
add support for Sketch 50 and fix text style matching

thierryc:
add any type to getSketchVersion


thierryc:
Already fixed in the : Add support for Sketch 50 and fix text style matching #290

thierryc:
outdated

Issue #294, opened 2018-04-21
Warning: Each child in an array or iterator should have a unique "key" prop


sogen:
_I am..._

| -------------------------------------------------------------------------------------------------
| Reporting a bug or issue
| -------------------------------------------------------------------------------------------------

**Expected behavior:**
Styleguide example builds without warnings or errors

**Observed behavior:**
Warnings start appearing from first run.
Console output:
```
console> Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `Palette`. See https://fb.me/react-warning-keys for more information.
    in Swatch (created by Palette)
    in Palette (created by Document)
    in view (created by View)
    in View (created by Section)
    in view (created by View)
    in View (created by Section)
    in Section (created by Document)
    in view (created by View)
    in View (created by Document)
    in Document
⛏  Built ./main.js in 1733ms
```

**How to reproduce:**
Clone to desktop
Run **Styleguide** example 

**Sketch version:**
49.2
Node v9.11.1
npm 5.6.0

**Please attach screenshots, a zip file of your project, and/or a link to your github project**
![screen shot 2018-04-20 at 6 02 25 pm copy](https://user-images.githubusercontent.com/2784323/39079016-b34ae210-44c7-11e8-975c-6e50418f69be.png)

mathieudutour:
it's in the styleguide example, one of the loops must have keys missing. Should be an easy PR up for grabs

csantiago132:
@mathieudutour I cloned repo (master) and ran `npm run test` and am getting some errors without touching code or anything. Maybe I did something wrong? 

```bash
Summary of all failing tests
 FAIL  __tests__/components/RedBox.js
  ● <RedBox /> › renders simple errors

    expect(value).toMatchSnapshot()

    Received value does not match stored snapshot 1.

    - Snapshot
    + Received

    @@ -42,11 +42,11 @@
               "color": "white",
               "fontFamily": "Monaco",
               "marginTop": 20,
             }
           }>
    -      invariant
    +      invariant (/Users/carlos/Dropbox/_Local/Github/public
         </text>
         <text
           name="Text"
           resizingConstraint={undefined}
           style={
    @@ -54,11 +54,11 @@
               "color": "white",
               "fontFamily": "Monaco",
               "marginTop": 20,
             }
           }>
    -      Object.<anonymous>
    +      Object.<anonymous> (/Users/carlos/Dropbox/_Local/Github/public
         </text>
         <text
           name="Text"
[...]

PR #311, opened 2018-05-02
Feature/docs styleguide

tbrasington:
Fix for #294 

The removal for the comma on line 9, I didn't explicitly do it but then nor can I commit it back. Not sure what is going on there.


jaridmargolin:
@tbrasington there is a pre-commit hook which runs prettier on staged files. Comma is intended to be removed.

Additionally, thanks for the quick patch :) LGTM

Issue #313, opened 2018-05-03
Automate generation of eslint globals & flow definitions


jaridmargolin:
This may not be deemed high priority, but it feels like there is unnecessary friction in managing both the eslint globals and flow definitions. I have a few questions that I believe would need to be answered in order to move forward:

### `skpm/eslint-config-sketch` vs `@jongold/eslint-config-sketch`

What are the advantages of using one over the other? Was work just being done in parallel or are there intentional differences between the two? 

I do notice that `@jongold/eslint-config-sketch` includes a few [non-sketch globals](https://github.com/jongold/eslint-config-sketch/blob/master/nonSketchGlobals.js).

What would need to be done in order to converge on a single lib?

### Non-sketch globals: `react-sketchapp` vs `@jongold/eslint-config-sketch`

Is there any rule of thumb of when to add globals directly in `react-sketchapp/.eslintrc` vs when to add rules within `@jongold/eslint-config-sketch`?

I think it would make more sense to manage all of these "non-sketch" globals in a single location.

### Can we automate creation of non-sketch globals?

It looks like all additional globals are coming from MacOS libs (with the exception of `context` and `log` which I believe are skpm specific?). Can we get these values automatically by utilizing `classdump` (similar to what is done in [sketch-headers](https://github.com/skpm/sketch-headers))?

### Can we automate creation of flow definitions?

Similar to the automation techniques used for generating eslint-config... can we dynamically create flow definitions from [sketch-headers](https://github.com/skpm/sketch-headers)?

---

@mathieudutour @jongold

jongold:
Feel free to deprecate my config. I think it has a lot of legacy
cocoascript stuff that we don't need anymore!

mathieudutour:
IIRC there is already a eslint config on the skpm org that is auto-generated. It should be just a matter of switching the dependencies

PR #314, opened 2018-05-04
cleanup and bump dev dependencies

mathieudutour:
- [x] run prettier across the entire codebase
- [x] bump dev dependencies to their latest version
- [x] add flow types for `sketch-constants` and `sketchapp-json-plugin`
- [x] update snapshots (because of the update of jest)
- [x] change the pre-commit hook to run prettier on tests and examples as well
- [x] use `eslint-config-sketch` instead of `@jongold/eslint-config-sketch` (connect #313 )


src/jsonUtils/computeYogaTree.js
```diff
@@ -29,7 +29,6 @@ const walkTree = (tree: TreeNode, context: Context) => {
 
   return node;
 };
-const treeToNodes = (root: TreeNode, context: Context): yoga.NodeInstance =>
-  walkTree(root, context);
+const treeToNodes = (root: TreeNode, context: Context): yoga.Yoga$Node => walkTree(root, context);
```
jaridmargolin: Nothing wrong here... I have just been trying to figure out what this syntax is: `Yoga$Node`. I have seen it elsewhere in the code within `React$Component` & `React$Element`

__tests__/components/RedBox.js
```diff
@@ -1,21 +1,16 @@
...
 import renderer from 'react-test-renderer';
 import RedBox from '../../src/components/RedBox';
 
 describe('<RedBox />', () => {
-  it('renders simple errors', () => {
-    try {
-      invariant(false, 'THIS IS AN ERROR');
-    } catch (e) {
-      const tree = renderer.create(<RedBox error={e} />).toJSON();
+  it.skip('renders simple errors', () => {
```
jaridmargolin: Can we mock the error?
Or... split this into 2 tests where each test case is wrapped in a conditional to test the node version and only execute if supported?

.flowconfig
```diff
@@ -30,14 +30,16 @@
...
+.*/lib
 .*/examples/.*
 
 [include]
 
 [libs]
 flow-typed/fbjs.js
+flow-typed/sketch-constants.js
 flow-typed/sketch.js
+flow-typed/sketchapp-json-plugin.js
```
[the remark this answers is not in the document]
mathieudutour: ah yes, I didn't know (not very familiar with flow)!

src/components/Artboard.js
```diff
@@ -9,28 +9,23 @@ const propTypes = {
...
-      ])
+      PropTypes.oneOfType([PropTypes.shape({ ...ViewStylePropTypes }), PropTypes.number]),
     ),
     PropTypes.number,
   ]),
   name: PropTypes.string,
   children: PropTypes.node,
 };
 
+// $FlowFixMe
```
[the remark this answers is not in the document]
mathieudutour: flow now requires `React.Component` to define the props it accepts. So we would need to duplicate the info in the prop-types. I'm not keen to duplicate codes and I don't want to lose the runtime check that prop-types bring. So I ignored it for now. Not sure what we should do.

src/jsonUtils/computeYogaTree.js
mathieudutour: it's the type definition that comes with yoga: https://github.com/facebook/yoga/blob/master/javascript/sources/entry-common.js#L123

src/components/Artboard.js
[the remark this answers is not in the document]
jaridmargolin: Not necessary in this PR, but what about https://github.com/brigand/babel-plugin-flow-react-proptypes?

src/jsonUtils/computeYogaTree.js
jaridmargolin: Ahhh. So it is just a convention. Awesome.

src/components/Artboard.js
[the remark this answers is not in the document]
mathieudutour: I thought about it as well. I read somewhere that it didn't really work but worth a try I guess.

__tests__/components/RedBox.js
mathieudutour: yeah, I'll try to mock. I don't even have the same stack trace locally than on travis...

mathieudutour:
and yes, we always squash the PRs

PR #317, opened 2018-05-08
version 2.0.0

mathieudutour:


jaridmargolin:
@mathieudutour I know you want to get a release out due to Sketch v50, however I was working through a few additional changes that would be considered "breaking".

Specifically I had toyed with the idea of adjusting the return values of `makeSymbol` in order to support some new workflows (I want to control injecting symbols myself). I may be able to work around breaking changes by introducing new methods rather than repurposing existing methods, but really I would say these issues are to early in development to discuss in detail. What are your thoughts around how often it would acceptable to push "breaking" changes that require major bump?

mathieudutour:
I don't mind publishing breaking changes as long as their usefulness outweigh the cost of updating :)

README.md
```diff
@@ -16,7 +16,7 @@ Open a new Sketch file, then in a terminal:
 
 ```bash
 git clone https://github.com/airbnb/react-sketchapp.git
-cd react-sketchapp/examples/basic-setup && npm install
+cd react-sketchapp/examples/basic-setup && npm installLobby
```
jaridmargolin: Is this right?

jaridmargolin:
I don't mind giving this the green light... but wondering if you are waiting for feedback from airbnb folks before releasing a new major bump.

PR #321, opened 2018-05-09
Remove `nvm install-latest-npm` from `.travis.yml` to fix node6 failure

jaridmargolin:
This may have been a perfect storm of events, but at some point today we started seeing failures on Travis CI when running against node 6.

Specifically [the error](https://travis-ci.org/airbnb/react-sketchapp/jobs/376605617) was occurring during the `postinstall` step of `spawn-sync` (which is a dependency of a dependency). As I began digging into debug I noticed some odd behavior:

Reviewing the last PR which passed I saw the output (slightly modified for brevity):
```
$ node --version
v6.14.2
$ npm --version
5.1.0
$ nvm --version
0.33.11
$ nvm install-latest-npm
* npm upgraded to: v5.1.0
```

And in the current failing tests I saw:
```
$ node --version
v6.14.2
$ npm --version
3.10.10
$ nvm --version
0.33.11
$ nvm install-latest-npm
* npm upgraded to: v6.0.0
```

Specifically two things stood out:
1. For some reason the passing tests had incorrectly installed npm v5.1.0 ([node 6.14.2 should install 3.10.10 by default](https://nodejs.org/en/download/releases/))
2. nvm doesn't appear to be bumping the npm version in the passing tests... this may be an oversimplification of what is actually happening behind the scenes.

The important takeaway here, is that the passing tests utilize npm v5.1.0 and the failing tests are utilizing npm v6.0.0. 

If I had more time I would attempt to track down why npm v6.0.0 is throwing an error when installing `spawn-sync` on node v6.14.1. For now however, to allow us to merge a few required PRs before releasing v2.0.0, the most straightforward approach seemed to be removing `nvm install-latest-npm` from our CI config.

cc @mathieudutour @ljharb

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,6 @@ node_js:
   - '6'
   - '8'
   - '10'
-before_install:
-  - nvm install-latest-npm
 before_script:
   - npm prune
 script:


ljharb:
@mathieudutour I'm confused why this was merged; installing the latest supported npm in CI is a best practice. The npm version won't necessarily be bumped; it depends on the node version, and what npm ships with it, and what the latest npm version that node version supports is.

If npm 6 is breaking with a transitive dep of yoga-layout, then there's a bug with yoga-layout that needs fixing, asap - since users will be using npm 6.

mathieudutour:
it's only breaking with node 6. I'd say it's pretty unlikely that users will use node 6 with npm 6.

Is it really a best practice? Ideally, we should test with the default npm version that is bundled with node and the latest. But I'd say that testing the default npm version is more important because it's what most people are using (especially designers that want to try react-sketchapp)?

ljharb:
I just realized the same thing; https://twitter.com/ljharb/status/994103496575107072

Definitely let's get the bug fixed in npm first, and then we can revert this later.

ljharb:
k, found a workaround. I set up travis-ci to have `NPM_CONFIG_AUDIT=false`, and I'll remove it when the next npm 6 comes out. This PR can now be safely reverted.

jaridmargolin:
Thank you @ljharb and @mathieudutour for taking this and running with it...

PR #324, opened 2018-05-12
Refactor SymbolInstanceRenderer `extractOverrides` method

jaridmargolin:
For the most part, this PR is a little superficial. I had these stashed locally on my computer from when I was attempting to work through some bugs with symbols. I found some of the code difficult to grok / inconsistent so I slowly started making some transformations.

I think the best way to review this PR would be to review each individual commit.

### DIFF
--- a/src/renderers/SymbolInstanceRenderer.js
+++ b/src/renderers/SymbolInstanceRenderer.js
@@ -26,8 +26,19 @@ const overrideProps = (layer: SketchLayer): Override => ({
   name: layer.name,
 });
 
+const removeDuplicateOverrides = (overrides: Array<Override>): Array<Override> => {
+  const seen = {};
+
+  return overrides.filter(({ objectId }) => {
+    const isDuplicate = typeof seen[objectId] !== 'undefined';
+    seen[objectId] = true;
+
+    return !isDuplicate;
+  });
+};
+
 const extractOverridesHelp = (subLayer: any, output: any) => {
-  if (subLayer._class === 'text' && !output.some(r => r.objectId === subLayer.do_objectID)) {
+  if (subLayer._class === 'text') {
     output.push(overrideProps(subLayer));
     return;
   }
@@ -78,7 +89,7 @@ const extractOverridesHelp = (subLayer: any, output: any) => {
 const extractOverrides = (subLayers: any) => {
   const output = [];
   subLayers.forEach(subLayer => extractOverridesHelp(subLayer, output));
-  return output;
+  return removeDuplicateOverrides(output);
 };
 
 class SymbolInstanceRenderer extends SketchRenderer {
--- a/src/renderers/SymbolInstanceRenderer.js
+++ b/src/renderers/SymbolInstanceRenderer.js
@@ -23,6 +23,9 @@ import { makeImageDataFromUrl } from '../jsonUtils/hacksForJSONImpl';
 const findInGroup = (layer: ?SketchLayer, type: string): ?SketchLayer =>
   layer && layer.layers && layer.layers.find(l => l._class === type);
 
+const hasImageFill = (layer: SketchLayer): boolean =>
+  !!(layer.style && layer.style.fills && layer.style.fills.some(f => f.image));
+
 const extractOverridesHelp = (subLayer: any, output: any) => {
   if (subLayer._class === 'text' && !output.some(r => r.objectId === subLayer.do_objectID)) {
     output.push({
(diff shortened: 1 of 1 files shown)


src/renderers/SymbolInstanceRenderer.js
```diff
@@ -20,75 +20,70 @@ import { makeImageDataFromUrl } from '../jsonUtils/hacksForJSONImpl';
...
-  if (subLayer._class === 'text' && !output.some(r => r.objectId === subLayer.do_objectID)) {
-    output.push({
-      type: 'text',
-      objectId: subLayer.do_objectID,
-      name: subLayer.name,
-    });
-    return;
+const findInGroup = ({ layers = [] } = {}, type) => layers.find(l => l._class === type);
+
+const hasImageFill = ({ style = {} } = {}) => style.fills && style.fills.some(f => f.image);
```
mathieudutour: I think that's actually wrong (I know it's not your bug but I just noticed it now):
- if the fill is disabled, it should not appear in the overrides
- there can be an image set but it could still be a color. We need to check the type of the fill

jaridmargolin:
Ahhh... Thats what I get for trying to push up a stashed changes without running checks locally... Wonder if flow can be added to precommit hook.

mathieudutour:
shouldn't be hard to add yes

src/renderers/SymbolInstanceRenderer.js
```diff
@@ -36,7 +36,7 @@ const extractOverridesHelp = (subLayer: any, output: any) => {
     return;
   }
 
-  if (subLayer._class === 'group' && subLayer.layers) {
```
jaridmargolin: This can be removed because I now handle checking for layers within the `findInGroup` method.

src/renderers/SymbolInstanceRenderer.js
```diff
@@ -70,12 +70,7 @@ const extractOverridesHelp = (subLayer: any, output: any) => {
     });
   }
 
-  if (
-    (subLayer._class === 'shapeGroup' ||
-      subLayer._class === 'artboard' ||
-      subLayer._class === 'group') &&
-    subLayer.layers
-  ) {
+  if (subLayer.layers) {
```
jaridmargolin: When I removed this a few weeks ago, I was fairly confident? But I will back this out because I am not 100% sure what my reasoning was.

src/renderers/SymbolInstanceRenderer.js
jaridmargolin: Ok.. I'll give this a look.

jaridmargolin: @mathieudutour a few notes after giving this more thought... 
1. This method is technically correctly doing what it says. The two concerns you brought up would most likely be addressed below where `hasImageFill` is called.
2. If we look at where this method is actually called, we get a little more context. It is first looking for a `ShapeGroup`, and then checking for if `hasImageFill`. To me this implies the current intent is actually to find images rendered specifically by `react-sketchapp`.
3. If it is in fact only looking for images rendered by `react-sketchapp`, do we need to account for the use cases you described above? Is it possible for a `react-skechapp` rendered image to appear in these two states?

If we want to change those decisions, I would prefer to kick them to a new issue. Mostly because I am not terribly interested in solving them right now 😬

jaridmargolin:
@mathieudutour I think this is ready for a second pass review.

src/renderers/SymbolInstanceRenderer.js
mathieudutour: Fair enough haha.  

I was thinking the same thing but I wasn’t sure if the function was used when getting an existing symbol as well. If that’s the case, then it’s not only react-sketchapp components that we deal with

src/renderers/SymbolInstanceRenderer.js
```diff
@@ -20,75 +20,79 @@ import { makeImageDataFromUrl } from '../jsonUtils/hacksForJSONImpl';
...
-      name: subLayer.name,
-    });
-    return;
+const findInGroup = (layer: ?SketchLayer, type: string): ?SketchLayer =>
+  layer && layer.layers && layer.layers.find(l => l._class === type);
+
+const hasImageFill = (layer: SketchLayer): boolean =>
+  !!(layer.style && layer.style.fills && layer.style.fills.some(f => f.image));
+
+const layerProps = (layer: SketchLayer): Object => ({
```
mathieudutour: Maybe a more explicit type than just `Object`? It’s used in a lot of places here

jaridmargolin: Yep... So I was torn here. There should be a type for Override, but when I went to add it, I saw that it previously existed and was removed: https://github.com/airbnb/react-sketchapp/commit/fae2bf7e9924d4823d285c71acdb14e4c4db2fe0#diff-cca6d1c7f1f3d777e20240c464d8f1b8L9

Although It does seem that the original removal was done purely to allow for release.

Issue #327, opened 2018-05-14
Why is the symbols implementation tied to mastersNameRegistry?


jaridmargolin:
I've been searching the codebase and reviewing the symbols implementation, and I keep coming to the conclusion that I must be missing something obvious. I am trying to determine why exactly the `mastersNameRegistry` exists and why it acts as the backbone of the symbols implementation.

Every symbol has a `symbolID`. It seems as if we could remove a ton of cruft and unnecessary `find` operations, if we instead had a `symbolsRegistry` which was keyed by the `symbolID` rather than the `masterName`.

mathieudutour:
I don't think there is any reason except that the exposed API to get an existing symbol is to query it by its name.

PR #329, opened 2018-05-14
Replace `mastersNameRegistry` with `symbolsRegistry`

jaridmargolin:
This builds off of the question posed in #327. There are several reasons why this change is beneficial:

1. Optimized internals. Right now querying items by `symbolID` requires iterating over the `mastersNameRegistry` and searching for a matching ID. This compounds when you realize this iteration occurs within some recursive functions inside `SymbolInstanceRenderer`. This PR ensures all queries made within our internals are done in constant time.
2. Consistent internal API. Right now there are two methods used internally for looking up symbolMasters from the registery, `getSymbolMasterByName` and `getSymbolMasterById`. This leads to possible misuse (see 1 above) and confusion. In short, less methods in use = less developer overhead.
3. I prefer it 😂Hopefully others agree that this is a more intuitive solution.

**Additional Notes**:

A part of me wanted to simplify `getSymbolID` to:

```js
const getSymbolID = (masterName: string): string => {
  const symbolMaster = getSymbolMasterByName(masterName)
  return symbolMaster ? symbolMaster.symbolID : generateID();
};
```

but I thought I'd keep this PR more self contained and address this in a possible subsequent PR.

### DIFF
--- a/src/renderers/SymbolInstanceRenderer.js
+++ b/src/renderers/SymbolInstanceRenderer.js
@@ -2,14 +2,14 @@
 import SketchRenderer from './SketchRenderer';
 import { makeSymbolInstance, makeRect, makeJSONDataReference } from '../jsonUtils/models';
 import type { ViewStyle, LayoutInfo, SketchLayer, TextStyle } from '../types';
-import { getSymbolMasterByName, getSymbolMasterById } from '../symbol';
+import { getSymbolMasterById } from '../symbol';
 import { makeImageDataFromUrl } from '../jsonUtils/hacksForJSONImpl';
 
 type Override = {
   type: string,
   objectId: string,
   name: string,
-  symbolId?: string,
+  symbolID?: string,
   width?: number,
   height?: number,
 };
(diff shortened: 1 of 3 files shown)


src/renderers/SymbolInstanceRenderer.js
```diff
@@ -71,7 +71,7 @@ const extractOverridesReducer = (
   if (layer._class === 'symbolInstance') {
     return overrides.concat({
       ...overrideProps(layer),
-      symbolId: layer.symbolID,
```
mathieudutour: is that a breaking change?

jaridmargolin: So this is actually misleading due to some inconvenient naming conventions which I plan to address... These are actually the `overridables` and not the overrides. The `overridables` are only used internally so it shouldn't be a problem.

Elsewhere in this PR however, I removed the `masterName` static prop on the created `SymbolInstance`. This would probably be considered a breaking change. There is no reason I can't put it back in.

I have however hit a few restraints with the current symbols implementation that will REQUIRE breaking changes.

mathieudutour: `masterName` is not a prop right? just a class field I believe. Let's put it back and I'll merge the PR

jaridmargolin: It was a prop on the `symbolinstance` but that is only used internally so I believe it is safe to remove there.

It was additionally set as a static property on the `React.Component` returned in the factory used to generate user facing `SymbolInstance`. I have just added it back here as there is a chance someone was relying upon it being there.

PR #334, opened 2018-05-22
Remove all commonjs style exports

jaridmargolin:


src/components/Svg/index.js
```diff
@@ -1,23 +1 @@
...
-module.exports.Polyline = require('./Polyline');
-module.exports.RadialGradient = require('./RadialGradient');
-module.exports.Rect = require('./Rect');
-module.exports.Stop = require('./Stop');
-module.exports.Symbol = require('./Symbol');
-module.exports.Text = require('./Text');
-module.exports.TextPath = require('./TextPath');
-module.exports.TSpan = require('./TSpan');
-module.exports.Use = require('./Use');
+export { default } from './Svg';
```
mathieudutour: Is it the es6 syntax or does it only works because we compile it (and it adds a named default export)?

(I’m just curious, haven’t stumbled upon it before)

ljharb: It’s standard syntax; it’s currently the only way to re-export a default as a default.

PR #345, opened 2018-05-30
handle fixed position

mathieudutour:
because there is no scrolling, `fixed` is actually the same as `absolute`. Let's allow it!


mathieudutour:
Oh indeed, I didn't know about that. TIL

PR #348, opened 2018-05-31
run e2e tests on CI

mathieudutour:
I needed to rework a bit some part because when running the tests, we only get an `MSDocumentData` instead of an `MSDocument`. That's actually fine because we only need that. Just need to make sure that we use the documentData instead of the document throughout the lib.

### DIFF
--- a/__tests__/skpm/basic.test.js
+++ b/__tests__/skpm/basic.test.js
@@ -2,6 +2,9 @@ import * as React from 'react';
 import * as sketch from 'sketch'; // eslint-disable-line
 import { render, View, Artboard, Text } from '../../src';
 
+// depending on where those tests run, we don't get the things,
+// eg. the context might be empty or there is no selected document
+// This make sure we always get something
 function getDoc(context, document) {
   return context.document || (sketch.getSelectedDocument() || document).sketchObject;
 }
--- a/__tests__/skpm/basic.test.js
+++ b/__tests__/skpm/basic.test.js
@@ -2,6 +2,10 @@ import * as React from 'react';
 import * as sketch from 'sketch'; // eslint-disable-line
 import { render, View, Artboard, Text } from '../../src';
 
+function getDoc(context) {
+  return context.document || sketch.getSelectedDocument().sketchObject;
+}
+
 const colorList = {
   Haus: '#F3F4F4',
   Night: '#333',
@@ -14,7 +18,7 @@ const colorList = {
 };
 
 test('should render a Page with a rectangle', (context) => {
-  const nativePage = context.document.currentPage();
+  const nativePage = getDoc(context).currentPage();
   // eslint-disable-next-line
   const Swatch = ({ name, hex }) => (
     <View
--- a/__tests__/skpm/basic.test.js
+++ b/__tests__/skpm/basic.test.js
@@ -2,8 +2,8 @@ import * as React from 'react';
 import * as sketch from 'sketch'; // eslint-disable-line
 import { render, View, Artboard, Text } from '../../src';
 
-function getDoc(context) {
-  return context.document || sketch.getSelectedDocument().sketchObject;
+function getDoc(context, document) {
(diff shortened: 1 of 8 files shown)


ljharb:
What only supports node 8, and why?

mathieudutour:
`@skpm/test-runner` only supports node 8+ and because it's using async/await and I don't want to compile it

ljharb:
Why not transpile it, or write in normal Promises instead of using the sugar?

mathieudutour:
I'm not sure why we are discussing it here. It's a test runner that only supports node 8+ that's about it. I'm not even sure why we are running the tests on different node versions since this library is compiled anyway...

ljharb:
The library being transpiled is irrelevant; it still has to be verified to run on supported node versions.

If a test runner won't work on an LTS version of node, then I don't think it's really usable on node projects.

mathieudutour:
>  it still has to be verified to run on supported node versions

This is not a node library so there are no supported node versions? What we could (and maybe should) test is different macOS versions and different Sketch versions.

> If a test runner won't work on an LTS version of node, then I don't think it's really usable on node projects

Again, that's not a node project.

I'm really not sure why we are discussing this. If you don't want e2e testing then fine, let's just drop it /shrug

ljharb:
I think e2e testing is great, and I don't want to drop it.

I was reacting to the comment about the tests failing on node 6.

If this isn't a node project, then that would mean node isn't required to be installed to use it - that doesn't seem like the case tho.

mathieudutour:
the only thing node is used for is installing the library. It doesn't run in node in any way.

ljharb:
(assuming you mean the whole project; not just the test runner) In that case, perhaps there's no value in running the tests on multiple node versions at all, and we can replace all three of 6/8/10 with `lts/*`?

__tests__/skpm/basic.test.js
```diff
@@ -2,6 +2,10 @@ import * as React from 'react';
 import * as sketch from 'sketch'; // eslint-disable-line
 import { render, View, Artboard, Text } from '../../src';
 
+function getDoc(context, document) {
```
jaridmargolin: Could you add a comment regarding what this function is doing, and why it is necessary?

src/symbol.js
```diff
@@ -32,13 +33,13 @@ const msListToArray = (pageList) => {
   return out;
 };
 
-export const getSymbolsPage = (document: any) => {
+export const getSymbolsPage = (document: SketchDocumentData) => {
```
jaridmargolin: 🌈

src/symbol.js
```diff
@@ -32,13 +33,13 @@ const msListToArray = (pageList) => {
...
   const pages = document.pages();
   const array = msListToArray(pages);
   return array.find(p => String(p.name()) === 'Symbols');
 };
 
-const getExistingSymbols = (document: any) => {
+const getExistingSymbols = (document: SketchDocumentData) => {
```
jaridmargolin: ✨

mathieudutour:
> Could you provide a little insight into why everything exists on document.documentData() rather than document?

`document` is an instance of `MSDocument` which is a subclass of `NSDocument`.
`document.documentData()` is an instance of `MSDocumentData` which is the root of our data model.

All the manipulations of the model happens on the MSDocumentData but there are some convenience methods on MSDocument that just forward it to the MSDocumentData.

For `react-sketchapp`, we only want to deal with MSDocumentData but we can receive an MSDocument instance so we have to make sure we access its document data if that's the case.

jaridmargolin:
Would it then make sense to rename occurrences of `document` -> `documentData`? Likewise, would it make sense to rename the helper methods:

`getDocumentFromContainer` -> `getDocumentDataFromContainer`
`getDocumentFromContext` -> `getDocumentDataFromContext`

Definitely a little verbose, but removes any possible confusion.

PR #350, opened 2018-06-02
support transforms

### LINKED ISSUE #14: Handle transforms
There's some complexity in this -

- Handle rotations with native Sketch rotations
- otherwise, rasterize the transform-containing layers & apply CGTransforms to them

mathieudutour:
fix #14 

```js
const Swatch = ({ name, hex }) => (
  <View
    name={`Swatch ${name}`}
    style={{
      height: 96,
      width: 96,
      margin: 4,
      backgroundColor: hex,
      padding: 8,
      transform: 'scale(-1, 1) rotate(30) translate(10, 10)', // <---- 💥
    }}
  >
    <Text name="Swatch Name" style={{ color: textColor(hex), fontWeight: 'bold' }}>
      {name}
    </Text>
    <Text name="Swatch Hex" style={{ color: textColor(hex) }}>
      {hex}
    </Text>
  </View>
);
```

screen shot 2018-06-01 at 21 05 52

Only a few transforms are supported:
- translations
- rotations should be ok
- scales when using `1` or `-1`
- a healthy mix of those

3D transforms don't work. There are probably bugs when changing the transform origin.

There are tons of edge cases that I haven't tested.


weaintplastic:
Hell yeah! Great work @mathieudutour

jaridmargolin:
> There are tons of edge cases that I haven't tested.

Would you like to see this merged as is (assuming yes as a PR has been opened)? Not sure what type of balance this project wants to strike. Is implementing something better than nothing? Or would we prefer to have a fully baked solution in order to avoid end user confusion?

I can give this a set of eyes from a coding standpoint, but I am probably not the right person to speak on philosophy of what should / should not get merged.

mathieudutour:
I'm not sure either. I probably won't have time to work on that a lot more (especially because there are tons of edge cases). So I'd say it's easier to get it out there and fix the bug reports afterwards?

jaridmargolin:
@ljharb Do you mind weighing in quickly with whether or not you agree with merging changes of this nature (incomplete / not fully tested, but still providing value to a certain subset of users)?

thecalvinchan:
this is awesome, any updates on when we can see this merged?

thecalvinchan:
hmm, just tested for a bit yesterday and today. seems to not be working with rotates. will test more and follow up with more context

sml782:
Has the latest stable version supported transform now

Issue #369, opened 2018-07-05
Text Decoration Underline is not working on master


thecalvinchan:
👋 Hello! Thanks for contributing. Please use the template that matches your intention  

_I am..._

| -------------------------------------------------------------------------------------------------
| Reporting a bug or issue
| -------------------------------------------------------------------------------------------------

**Expected behavior:**

Using `styled-components/primitives` with

```
const Link = styled.Text`
  text-decoration: underline
`
```

should render text with underline

**Observed behavior:**

Throws a JSON Parse syntax error for unexpected token '}'

```
makeTextLayer@/Users/calvinchan/airswap/component-library/plugin.sketchplugin/Contents/Sketch/styleguide.js:35247:73
renderBackingLayers@/Users/calvinchan/airswap/component-library/plugin.sketchplugin/Contents/Sketch/styleguide.js:36270:43
flexToSketchJSON@/Users/calvinchan/airswap/component-library/plugin.sketchplugin/Contents/Sketch/styleguide.js:33653:51
map@[native code]
flexToSketchJSON@/Users/calvinchan/airswap/component-library/plugin.sketchplugin/Contents/Sketch/styleguide.js:33655:60
renderContents@/Users/calvinchan/airswap/component-library/plugin.sketchplugin/Contents/Sketch/styleguide.js:35326:44
map@[native code]
render@/Users/calvinchan/airswap/component-library/plugin.sketchplugin/Contents/Sketch/styleguide.js:35395:22
/Users/calvinchan/airswap/component-library/plugin.sketchplugin/Contents/Sketch/styleguide.js:78583:65
__skpm_run@/Users/calvinchan/airswap/component-library/plugin.sketchplugin/Contents/Sketch/styleguide.js:78699:17
__skpm_run@[native code]
  line: 35247,
  column: 73,
  sourceURL: '/Users/calvinchan/airswap/component-library/plugin.sketchplugin/Contents/Sketch/styleguide.js' }
```

**How to reproduce:**

Use `styled-components` example but change the dependency of `react-sketchapp` to point to master. Then add a `text-decoration: underline` attribute to any of the styeld components

**Sketch version:**

50.2

**Please attach screenshots, a zip file of your project, and/or a link to your github project**

PR #370, opened 2018-07-05
Fixed textDecoration missing value bug

thecalvinchan:
Fixes https://github.com/airbnb/react-sketchapp/issues/369

The issue was with the previous version of the ternaries, passing in a `textDecoration` that satisfied `NSUnderline` wouldn't satisfy `NSStrikethrough`, leaving the value of `NSStrikethrough` as undefined.

When passing into the JSON function later on, it would just leave the value blank, causing a parse error


thecalvinchan:
cool, fixed.

PR #382, opened 2018-09-20
Add support for paragraph spacing

lessthanzero:
Added a very basic support for the paragraph spacing. Also checked the creation text styles.
With @adryweber


mathieudutour:
`paragraphSpacing` isn't supported in react-native. What would be the use case of using `paragraphSpacing` instead of just using 2 Text layers?

mathieudutour:
Ah I see, make sense

PR #400, opened 2018-11-13
ci: test Node.js 6, 8, 10 and 11

DanielRuf:
LTS are 6, 8 and 10. `lts/*` only uses the last LTS release, which would be 10 now. See https://github.com/nodejs/Release/blob/master/README.md

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,10 @@ cache:
 notifications:
   email: false
 node_js:
-  - 'lts/*'
+  - 'node'
+  - '10'
+  - '8'
+  - '6'
 before_install:
   - brew update
   - brew cask install sketch # install Sketch


.travis.yml
```diff
@@ -8,7 +8,10 @@ cache:
 notifications:
   email: false
 node_js:
-  - 'lts/*'
+  - 'node'
```
ljharb:
```suggestion
  - '11'
```
there’s no need to have ci start failing randomly whenever node 12 comes out :-)

DanielRuf: Why not? Who then keeps an eye on the CI matrix to bump it to 12? Also `node` is recommended and used in many projects. So even if you forget to add further LTS versions when a new stable branch is released, you always have the CI running against the current stable release which catches any new issues (many update early and 11 is a shortliving branch like every odd release branch).

If this is an issue for you, we can add `node` to `allow_failures`.

As most do not check the CI status, you will get at least a green status but would have to check the details page if it failed.

ljharb: The same people who keep an eye on semver-major dependency updates, which don’t magically run on regular ci. Regular CI isn’t the place to just “hope” that after an update happens, someone notices a failure - that’s what things like greenkeeper are for.

Adding it to allowed failures should work fine tho, if you think this is valuable.

DanielRuf: > Adding it to allowed failures should work fine tho, if you think this is valuable.

Ok, I'll add a commit tomorrow probably for `allow_failures`.

Issue #409, opened 2019-01-29
Symbol overrides don't work


lessthanzero:
I am trying to use standard override example with the most recent `react-sketchapp` version and symbol overrides don't work, returning no errors and just a master symbol instance with default sub-symbol.

**Sketch version:**
52.6

lessthanzero:
@mathieudutour any news?

mathieudutour:
haven't had time to look into it no

lessthanzero:
Kinda found the solution with #412

PR #412, opened 2019-02-12
Fix symbol overrides behavior in Sketch 52+

lessthanzero:
Fixing #409 

Seems that overrides were broken due to the usage of an optional argument (max Sketch version or smth like that) in `fromSJSONDictionary`.

Kudos to @b2whats.


mathieudutour:
I don't think that's the right fix, there is a reason why the version is passed

mathieudutour:
the right fix is here: #413

Issue #420, opened 2019-03-30
Exporting `Svg` components to align with react-native-svg


saschazar21:
_I am..._

| -------------------------------------------------------------------------------------------------
| Requesting a new feature
| -------------------------------------------------------------------------------------------------

**Proposal/Feature-request:**

To provide a seamless integration possibility between `react-sketchapp` and `react-native-svg`, it should be possible to use the same export structure for `Svg` components as `react-native-svg` does (e.g. to support `@svgr/core`'s native parsing possibilities).

**Current situation:**

```jsx
import { Svg } from 'react-sketchapp';

export const CircleIcon = (props) => (
  <Svg viewBox="0 0 24 24" {...props}>
    <Svg.Circle cx={12} cy={12} r={1} />
  </Svg>
);
```

**`react-native-svg` approach:**

```jsx
import { Svg, Circle } from 'react-sketchapp';

export const CircleIcon = (props) => (
  <Svg viewBox="0 0 24 24" {...props}>
    <Circle cx={12} cy={12} r={1} />
  </Svg>
);
```

mathieudutour:
The difference is that `react-native-svg` is namespaced. You wouldn't expect `Circle` to be exported on `react-native` but `Svg` might make sense.

You can easily do 
```
import { Svg } from 'react-sketchapp';
const { Circle } = Svg;
```
tho

saschazar21:
Yes I'm completely with you on this one.

If I may suggest a different solution: why not exporting the `Svg` members in the `lib/components/Svg/index.ts` file? Therefore the namespace still makes sense and it would be possible to export each member without the need of a destructuring operation of the imported `Svg` class.

The issue I see with the destructuring option you mentioned above, is that it prevents tree-shaking and as well auto-fixing imports by `tslint`.

I might as well supply a PR for further consideration, if that's okay for you. Otherwise I'm also okay with closing this issue.

mathieudutour:
Happy to look at a PR yes

PR #422, opened 2019-04-02
Export Svg components in the Svg/index.js file

saschazar21:
This PR exports all `Svg` components (`Circle`, `Rect`, etc...) from the `lib/components/Svg/index.js` file.

This not only supports tree-shaking, it also improves the compatibility towards the `react-native-svg` API (especially when using tools like [SVGR](https://github.com/smooth-code/svgr) in combination with TypeScript).

Fixes: https://github.com/airbnb/react-sketchapp/issues/420


package.json
```diff
@@ -43,6 +44,7 @@
     "@skpm/sketchapp-json-plugin": "^0.2.0",
     "airbnb-prop-types": "^2.10.0",
     "error-stack-parser": "^2.0.0",
+    "gitbook-plugin-github": "^2.0.0",
```
mathieudutour: That shouldn't be here

saschazar21: Interestingly this dependency snuck in when I was running `yarn docs:watch`. Will remove it, however I can't guarantee it's not coming back some time in the future at someone else's place.

I think it's caused by this line in the `book.json`: https://github.com/airbnb/react-sketchapp/blob/master/book.json#L8

saschazar21: fixed in https://github.com/airbnb/react-sketchapp/pull/422/commits/8758f3783130a9138b1e40f4bd8dbe0ef6fcd5a6

mathieudutour:
Thanks!

Issue #280, opened 2018-03-07
Add a new primitive Hotspot for prototyping support


taojoe:
👋 Hello! Thanks for contributing. Please use the template that matches your intention  

I am Joe, a programmer, who is interested this project. 

| -------------------------------------------------------------------------------------------------
| Requesting a new feature
| -------------------------------------------------------------------------------------------------

support link as sketch. sketch now support link, can we support it?

https://www.sketchapp.com/docs/prototyping/links

so the react-sketchapp can also support designer to do prototyping.

csantiago132:
I don't think it matters; React Sketchapp renders to sketch, then in sketch you'll have to connect the artboards between each other. As long as React Sketchapp supports the current version of sketch, it wont be a problem.

mathieudutour:
well we could create a `HotSpot` component and add a `flow` property on the `View` for example

taojoe:
@mathieudutour 

will Hotspot be in next release?

mathieudutour:
If someone create a PR to support it maybe

jongold:
This is a great idea. I don't have time to work on it, but I'd love to see a PR for it!

macintoshhelper:
Hi, I've developed a proof of concept here: https://github.com/macintoshhelper/react-sketchapp/pull/new/feat/hotspot .

I'll work on cleaning it up for a PR. Are there any thoughts/feedback on how the API should work? 

`destinationArtboardID` expects an artboard UUID, but we do not have this info while rendering, so I've created a new `generateCachedID` function that can be run before the render to pre-populate all the artboard IDs, this feels a bit messy though.

mathieudutour:
Thanks for taking this!

Regarding the `destinationArtboardID`, I think we can try to do a bit of magic: the `generateID` method generates stable object IDs given a name. So if we generate the IDs of the artboards using that (I don't remember if we do already), we can have the prop takes the name of the artboard (and generate the ID from it) or even the artboard component instance?
And we can have a second prop that takes an ID directly so that you can reference existing artboards (not generated by react-sketchapp).
What do you think?

macintoshhelper:
As far as I can tell, `e7` needs to be used to generate the same UUID for a given seed/name, `generateID` is returning different UUIDs:

```js
it('is unique', () => {
  expect(e7()).not.toBe(e7()); // This passes
});
it('same id for same seed', () => {
  expect(e7('test')).toBe(e7('test')); // This passes
});
[...]

PR #436, opened 2019-08-16
Prototype flows with HotSpot and Artboard ids

macintoshhelper:
- New `<HotSpot>` component
- New `id` and `isHome` prop for `<Artboard>` to use for prototypes

Includes tests and documentation.

> ref #280


src/renderers/ArtboardRenderer.js
```diff
@@ -18,7 +18,7 @@ export default class ArtboardRenderer extends SketchRenderer {
 
     return {
       _class: 'artboard',
-      do_objectID: generateID(),
+      do_objectID: props.id ? generateID(props.id, true) : generateID(props.name),
```
mathieudutour: I'm wondering if we need to introduce a new `id` prop 🤔 

I think just have the name and making it the "hardcoded" seed all the time for artboard should be good enough. Maybe we can prefix the name with `artboard:` to make it won't conflict with other layers.

What do you think?

src/jsonUtils/models.js
```diff
@@ -57,16 +57,18 @@ function generateIdNumber() {
...
-  if (seed) {
+  if (seed && !isHardcoded) {
     if (!previousSeeds[seed]) {
       previousSeeds[seed] = 0;
     }
     previousSeeds[seed] += 1;
 
     _seed = `${seed}${previousSeeds[seed]}`;
+  } else if (seed && isHardcoded) {
+    _seed = seed;
```
mathieudutour: even if it's hardcoded, I think we want to increment `previousSeeds`

src/components/HotSpot.js
```diff
@@ -0,0 +1,34 @@
...
+  };
+
+  static defaultProps = {
+    name: 'HotSpot',
+  };
+
+  render() {
+    const { children, flow, ...props } = this.props;
+    return (
+      <View injectedProps={{ flow }} {...props}>
```
mathieudutour: so I see that for a `HotSpot` layer, you generate a normal View with the `flow` property. 

Sketch has a notion of `HotSpot` which is separated from a normal layer. I understand why you'd want to generate a normal layer but then why do we need another layer type? Could we just add another prop to `View`?

__tests__/jest/jsonUtils/models.js
```diff
@@ -11,6 +11,12 @@ describe('generateID', () => {
   it('is unique', () => {
     expect(generateID()).not.toBe(generateID());
   });
+  it('seed generates different ID', () => {
+    expect(generateID('test')).not.toBe(generateID('test'));
+  });
+  it('hardcoded seed generates same ID', () => {
+    expect(generateID('test', true)).toBe(generateID('test', true));
+  });
```
mathieudutour: 👍 thanks for the tests

src/renderers/ArtboardRenderer.js
macintoshhelper: I implemented the new `id` prop incase someone has duplicate artboard names, or leaves the default name of `Artboard`. I was trying to avoid cases of duplicate UUIDs, but maybe in that case, there should be a console warning instead? The `id` prop should match the `flow.target` input also, so a layer display name might not necessarily be what someone would want to use in `flow={{ target: 'id-name' }}`.

Maybe I could change it to name though, and leave it as a random UUID if `name` is left as `Artboard`, and leave the `id` prop to override the name if someone wants?

src/components/HotSpot.js
macintoshhelper: True, yes. I actually ran into an issue where I was using my UI lib which wraps `react-primitives` and `styled-system` to turn `<View>` into `<Box>`, etc, and found myself using `<Box injectedProps={{ flow: {...} }}` which is messy. I'll move the `flow` prop to `<View>`.

Should `src/index.js` have a `HotSpot` export which is a reference to `<View>`? e.g.

```js
import View from './components/View';

const HotSpot = View;

module.exports = {
  ...
  View,
  HotSpot,
  ...
```

Might be messy though if people are importing `HotSpot` and it is removed in the future.

__tests__/jest/jsonUtils/models.js
macintoshhelper: 🙂  Just a thought, could it be worth having React tree to Sketch JSON test cases in the repo? Probably wouldn't work in a CI though.

src/renderers/ArtboardRenderer.js
macintoshhelper: Thoughts on this solution? I replace `hardcoded` with an `index` prop to retrieve an ID by index, this means if there's duplicate artboards with the same name, the prototype link will go to the first one.

I think it would be worth using `name`, but with `id` as an optional override if you want a different layer display name. 

`generateID` function:

```js
export function generateID(seed?: ?string, index?: number): string {
  let _seed: ?string;

  if (seed && !index) {
    if (!previousSeeds[seed]) {
      previousSeeds[seed] = 0;
    }
    previousSeeds[seed] += 1;

    _seed = `${seed}${previousSeeds[seed]}`;
  } else if (seed && index) {
    _seed = `${seed}${index}`;
  }

  return e7(_seed);
}
```

`ArtboardRenderer.js`

```js
...
do_objectID: generateID(props.id || props.name, 1),
...
```

mathieudutour: I was more thinking something like:
```js
export function generateID(seed?: ?string, hardcoded?: boolean): string {
  let _seed: ?string;

  if (seed) {
    if (!previousSeeds[seed]) {
      previousSeeds[seed] = 0;
    }
    previousSeeds[seed] += 1;

    if (!hardcoded) {
      _seed = `${seed}${previousSeeds[seed]}`;
    }
  }

  return e7(_seed);
}
```

and in `ArtboardRenderer.js`

```js
...
do_objectID: generateID(`artboard:${props.name}`, true),
...
```
(so that we are sure it doesn't conflict with other types of layers

src/components/HotSpot.js
mathieudutour: let's forget about `HotSpot` for now then

__tests__/jest/jsonUtils/models.js
mathieudutour: I'd say it's more interesting to have E2E tests: actually rendering the components and checking it's what we expect using the Sketch API (there are some already: https://github.com/airbnb/react-sketchapp/tree/master/__tests__/skpm)

src/renderers/ArtboardRenderer.js
macintoshhelper: I prefer having an `id` optional override, so a dev can use `flow={{ target: '/about:mobile' }}` and a designer/Sketch user would see `About | Mobile` in the layers, and using `index` instead of `hardcoded` should work nicer for duplicate artboard names/ids I think?

I can revert to your proposed `generateID` and remove the `id` prop if you'd prefer though. Good point on the `artboard:` prefix 👍

Updated code diff: 
https://github.com/airbnb/react-sketchapp/pull/436/commits/6d0d88708698008991566ad1d01044a253d3c228

__tests__/jest/jsonUtils/models.js
macintoshhelper: Ok, will leave it out for this PR I think, but will look into building some more E2E test cases soon. Would be really nice to have E2E tests + Sketch builds working in a Linux CI in the future.

src/renderers/ArtboardRenderer.js
mathieudutour: > so a dev can use flow={{ target: '/about:mobile' }} and a designer/Sketch user would see About | Mobile in the layers

but why would the dev not do `flow={{ target: 'About | Mobile' }}`? It's really Sketch specific anyway so I don't think it matters if it looks like a URL or not, right?

> using index instead of hardcoded should work nicer for duplicate artboard names/ids

But using an `index` hardcoded as `1` instead of just a boolean won't change anything and might actually conflict with an existing `previousSeeds` 🤔

mathieudutour: I think I'd rather not add a new prop for now so that we don't expand the API surface (a lot harder to remove stuff than adding it), but let's keep it in mind if it turns out to be a pain

macintoshhelper: 👍Applied your suggestions. I agree that accepting an `id` prop for `<Artboard>` is messy and could lead people to expect the same prop to work for other components (inconsistency).

Would you have any thoughts on exposing internal API access for things like the seed override to library developers (on the lines of dangerouslySetInnerHTML)?

For `react-sketchapp-router`, I'd like to be able to do `<Link to="/about">` for an artboard with name `"About | Mobile"`, but that should be resolvable with an abstraction of an internal map of paths and artboard names in the `<Link>` and `<Route>` components.

macintoshhelper: The thinking behind using `index` instead of `hardcoded` for `generateID`, was to handle the case of multiple artboards with the same name. As far as I can tell, having `hardcoded` as true for all artboards, means duplicate names result in duplicate object IDs.

With `generateID(target, 1),`, the target falls back to the first Artboard component in the tree.

macintoshhelper:
> 👍

Thanks! I'll start work on an initial release of `react-sketchapp-router` 🙂

PR #468, opened 2020-01-19
Merge pull request #468 from compositive/change-default-font-resolution

lordofthelake:
Mirrors the changes applied in https://github.com/Lona/node-sketch-bridge/pull/7 to the Sketch implementation.


src/jsonUtils/sketchImpl/findFontName.ts
```diff
@@ -3,7 +3,6 @@
 import hashStyle from '../../utils/hashStyle';
 import { TextStyle } from '../../types';
 import { FONT_STYLES } from '../textLayers';
-import { APPLE_BROKEN_SYSTEM_FONT } from '../../utils/constants';
```
mathieudutour: is this used anywhere else? otherwise let's remove it from there as well

PR #472, opened 2020-01-20
Pure image fetching implementation + async render

lordofthelake:


src/components/Text.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import { or } from 'airbnb-prop-types';
-import StyleSheet from '../stylesheet';
```
mathieudutour: Could you limit this PR to related changes? It's already huge without those cosmetic changes haha

lordofthelake: Sorry I got carried away 😅 I'll remove the less relevant parts

lordofthelake:
@mathieudutour Before I spend time debugging it, is `npm run test:e2e` supposed to work? I see it's commented out on `.travis.yml` and it's erroring out in a cryptic way for me, so before investing the time I wanted to figure out whether it's me or it was broken before.

```sh
npm run test:e2e

> react-sketchapp@3.0.5 test:e2e /Users/michelepiccirillo/Development/COMPOSITIVE/react-sketchapp
> skpm-test

Looking for the test files
 RUNS  basic
 RUNS  render-context
 RUNS  render-in-wrapped-object

Building the test plugin source
Compiling the test plugin
Running the tests
error Error while running the tests after build
Error: Command failed: "/Applications/Sketch.app/Contents/Resources/sketchtool/bin/sketchtool" run "/Users/michelepiccirillo/Development/COMPOSITIVE/react-sketchapp/node_modules/@skpm/test-runner/test-runner.sketchplugin" "plugin-tests" --without-activating

    at ChildProcess.exithandler (child_process.js:295:12)
    at ChildProcess.emit (events.js:210:5)
    at maybeClose (internal/child_process.js:1021:16)
    at Socket.<anonymous> (internal/child_process.js:430:11)
    at Socket.emit (events.js:210:5)
    at Pipe.<anonymous> (net.js:659:12) {
  killed: false,
  code: 1,
  signal: null,
  cmd: '"/Applications/Sketch.app/Contents/Resources/sketchtool/bin/sketchtool" run "/Users/michelepiccirillo/Development/COMPOSITIVE/react-sketchapp/node_modules/@skpm/test-runner/test-runner.sketchplugin" "plugin-tests" --without-activating'
}
```

lordofthelake:
@mathieudutour It needs still some polishing, updating the docs, and writing some new tests, but it should be functionally complete now. I would appreciate some feedback while I work to wrap the rest up.

src/components/Svg/TextPath.tsx
```diff
@@ -12,9 +12,10 @@ export default class TextPath extends React.Component<Props> {
   render() {
     if (!this.props.href || !this.props.href.match(idExpReg)) {
       // eslint-disable-next-line no-console
-      console.warn(
-        `Invalid \`href\` prop for \`TextPath\` element, expected a href like \`"#id"\`, but got: "${this.props.href}"`,
-      );
+      if (process.env.NODE_ENV !== 'production')
```
mathieudutour: any reason for this?

src/jsonUtils/makeSvgLayer/makeSvgLayer.ts
```diff
@@ -78,8 +77,13 @@ function makeLayerGroup(
...
 
-export default function makeSvgLayer(layout: LayoutInfo, name: string, svg: string) {
-  const svgModel = requireSvgModel();
+export default async function makeSvgLayer(
+  layout: LayoutInfo,
+  name: string,
+  svg: string,
+): Promise<FileFormat.Group> {
+  // Load the module only if it has been made available through another import.
+  const svgModel = (await import(/* webpackMode: "weak" */ '@lona/svg-model')).default;
```
mathieudutour: are you sure the comment stays when compiling the file?

src/renderers/SymbolInstanceRenderer.ts
```diff
@@ -114,8 +114,8 @@ export default class SymbolInstanceRenderer extends SketchRenderer {
 
     const overridableLayers = extractOverrides(masterTree.layers);
 
-    const overrides = overridableLayers.reduce(function inject(
-      memo: FileFormat.OverrideValue[],
+    const overrides = await overridableLayers.reduce(async function inject(
```
mathieudutour: wouldn't
```ts
const overrides = (await Promise.all(overridableLayers.map((
  memo: Promise<FileFormat.OverrideValue | undefined>[],
  reference: Override
) => memo.concat(async () => {
  ...
})
, [])).filter(x => !!x)
```
be faster and possibly more understandable?

src/utils/getImageDataFromURL.ts
```diff
@@ -1,15 +1,69 @@
...
+const fetchRemoteImage = async (url: string, { fetch }: PlatformBridge): Promise<Buffer> => {
+  const response = await fetch(url);
+  if (!response.ok) {
+    throw new Error(`${response.status}`);
+  }
+
+  const arrayBuffer = await response.arrayBuffer();
+  if (Buffer.isBuffer(arrayBuffer)) {
+    // skpm polyfill returns a Buffer instead of an ArrayBuffer
+    return arrayBuffer;
```
mathieudutour: let's put that in the bridge maybe?

src/utils/getSketchVersion.ts
```diff
@@ -1,6 +1,5 @@
-export function getSketchVersion(): number | 'NodeJS' {
-  if (typeof NSBundle !== 'undefined') {
-    return parseFloat(NSBundle.mainBundle().infoDictionary().CFBundleShortVersionString);
-  }
-  return 'NodeJS';
+export default function getSketchVersion(): string {
```
mathieudutour: any reason for this change?

src/jsonUtils/makeSvgLayer/makeSvgLayer.ts
lordofthelake: Yup, I checked. TS doesn't remove the comments unless you tell it to

src/utils/getImageDataFromURL.ts
lordofthelake: I'm not sure that this one is worth making an API for in the bridge, because in theory every implementation should be returning an `ArrayBuffer` and it's just an idiosyncrasy of the skpm one that it returns a `Buffer` directly instead.

What I could do here though is removing this check and using always `Buffer.from(arrayBuffer)`. For all the compliant implementations it will work as expected. For the skpm one, it will just copy the memory again, that doesn't do much harm; and if you feel like patching the polyfill at a later date, it would still work fine and become a little bit more efficient. What do you think?

src/utils/getSketchVersion.ts
lordofthelake: One thing that occurred to me is that with the bridge system, this package could be running also in the browser, so returning `"NodeJS"` would be misleading. 

Also, parsing versions as a float, in general, is discouraged: in case there are patch versions, you lose that information; but, more importantly, v1.10 becomes the same as v1.1; the direct naive comparison of floats would also bring the wrong results: v1.10 > v1.9, but 1.10 < 1.9, so naive code would do the wrong thing.

src/components/Svg/TextPath.tsx
lordofthelake: In general, especially if you're running this as part of a web service, you don't want your libraries spamming the log, moreover so if you have a structured logging system (and you're not fully in control of the inputs because you're processing user data). 

A lot of node libraries use something akin to [debug](https://www.npmjs.com/package/debug) for this purpose, but I didn't want to go too far with this. I'm open to a different approach here if you have other ideas.

lordofthelake:
> I think I like the `.sketch` approach. I forgot I added that to skpm haha. It's old enough that it shouldn't be too bad of a breaking change.

If you want I can remove the runtime checks altogether and just work with the static bundling behavior. For sure it would be better for bundle sizes.

> 
> But the PR is pretty hard to review, there are a lot of changes that aren't related. 

I know it got a lot bigger than it was supposed to be, but I started pulling a long thread here that went like this: 
1. need to work with fetch -> start making things async -> bubbled up to the top 
2. need to get fetch from somewhere -> add the bridge injection -> bubbled up

At this point, it was worth making it possible to bundle it also on other platforms, so I cleaned up the various require hacks, and part of that is moving the methods together so it can work with the static resolution.

The things that I could leave out would be the logging. Possibly also I could not rename the *SJON functions, but that would leave a stray `sketchImpl` directory behind.

> Additionally, there are tons of merge conflicts (I turned on the TS strict mode a couple of days ago, sorry)

That's not a problem, most changes are mechanical, so I'll fix that. It mostly conflicts because Prettier split the method signatures into more lines when they got longer. 70% of the changes in this PR are adding `async/await`, returning Promises and adding an additional argument, so it's not a hard job.

src/renderers/SymbolInstanceRenderer.ts
lordofthelake: Possibly. I haven't really checked what it does, since it was quite convoluted; I just applied mechanically promises where relevant. I'll look into it.

lordofthelake: I've tried giving a quick stab at this, but it's not as straightforward, because it has parts where it calls itself recursively with other `reduce`. 

Definitely it needs refactoring because it's horribly windy and convoluted, and probably could use being broken down in multiple more focused methods as well, but I'm not sure if this is the PR where it's best to tackle it.

src/sharedStyles/TextStyles.ts
```diff
@@ -1,6 +1,13 @@
 import { FileFormat1 as FileFormat } from '@sketch-hq/sketch-file-format-ts';
```
lordofthelake: This file is better reviewed side-by-side. The change is trivial, but the diff that git produces is horrible

lordofthelake:
I have fixed all the merge conflicts and some other minor adjustments. Sorry, I had to bring back `tsconfig.module.json`: without modules it's not possible to have webpack apply tree shaking. It's not as relevant for Sketch, but if the browser becomes a target, it's a good to have.

lordofthelake:
@mathieudutour I have reverted the changes to logging and the version handling, for now, to make the review easier. I'll re-submit those in separate, more focused, PRs later.

Let me know if there is anything else I can do to make this easier to merge.

lordofthelake:
@mathieudutour Friendly ping :)
I would appreciate some feedback on whether the overall direction is to incorporate this change at all or not, since we have a bunch of work downstream influenced by this. 

If affirmative, I will do the last round of polishing with docs & tests so that this can go in properly; otherwise, I'll just move on with a different approach for our code.

mathieudutour:
Sorry for the delay. Yeah, happy with how it's going 👍

macintoshhelper:
> @mathieudutour Before I spend time debugging it, is `npm run test:e2e` supposed to work? I see it's commented out on `.travis.yml` and it's erroring out in a cryptic way for me, so before investing the time I wanted to figure out whether it's me or it was broken before.
> 
> ```shell
> npm run test:e2e
> 

Hi @lordofthelake , just noticed this comment. I actually ran into the same issue on #516 . I'm working on fixing up the e2e tests, and adding e2e tests for all of the examples, with the view of integrating visual regression testing later on.

It appears to be a problem on Sketch's side with `sketchtool`, or with `skpm-test`. Importing the test plugin manually (`open node_modules/@skpm/test-runner/`, double clicking `test-runner.sketchplugin` and using `sketch-dev-tools` to see console logs works as a hacky fix for now.

Running `npm run test:e2e` in `git checkout b019845f28f2a53dc257d6ff16b2249841858242` should help figure out if this is an issue introduced by a Sketch update or not (e2e testing worked in that commit/git snapshot).

macintoshhelper:
@lordofthelake Hi, btw, I found what was breaking the e2e tests. Adding this to the `package.json` should fix it:

```json
{
// ...
  "dependencies": {
// ...
    "@skpm/test-runner": "^0.3.2",
  }
```

Ref: #516

yezz123 (approved):

lordofthelake:
This has been around for a while, although 50%+ of the changes proposed here have already been incorporated. I have cleaned up the rest of the changes and resolved the merge conflicts.

There should probably be a major version bump with this, since the switch to returning `Promise`s is not retro-compatible

lordofthelake:
The build is red for what I suspect is a dirty npm cache (https://github.com/nodejs/help/issues/2874). I would try to clean Travis' cache, but I don't have enough permissions to do so. Maybe @mathieudutour can help with this.

huynhdev24 (approved):

Issue #343, opened 2018-05-30
react-sketchapp-router


mathieudutour:
I'm wondering if it would be possible to create a router for react-sketchapp. It would render all the routes (Artboards) at once in a page and set up the links so that they can be viewed as a prototype.

It's just a fuzzy idea for now but it might just work. Bonus point if we can mimic the `react-router` API

cc @jaridmargolin

jaridmargolin:
Hmmm... maybe you could expand.

As I think through the implementation, `react-router` is really just path based conditional routing. Match this path, cool, render. I'm not entirely sure how this concept would carry over to Sketch 🤔

I could see, a `Link` component which has a `to` prop that accepts an `Artboard` name. It would be pretty straightforward to build? The only thought is that it doesn't quite fit the paradigm of a primitive. It would be a Sketch specific component.

macintoshhelper:
I've published a proof of concept repo for `react-sketchapp-router` here: https://github.com/macintoshhelper/react-sketchapp-router

It's using a fork of `react-sketchapp` with `<HotSpot>` for which there's an open PR: #436 

Would be great to have ideas/feedback on the API. `<Router>` `history.location` prop is hardcoded in the provider and is used for `<Artboard>` `isHome` prop.

I'm trying to think of a nice way of having mobile, tablet and desktop routes be rendered using it so that it can be used to test responsive designs quickly.

mathieudutour:
That's *amazing*! Exactly what I had in mind!

The `viewport` prop is a nice touch. What are the API differences with `react-router`? Eg. could we create a `react-router-primitives` that requires `react-router` or `react-sketchapp-router` depending on the platform you renders to without any change in user code? That would make an amazing demo

macintoshhelper:
Thanks! I've updated the repo a bit with a new `<SketchRouter>` and `<Switch>` component. A `react-router-primitives` lib would be great! For now, I'm using a shared `routes.js` file and separate entry-points for each render platform.

> @mathieudutour 
Have you had a chance to look at my updates to the #436 PR? I've implemented your suggestions and left a comment on the `generateID` function.

Rough API spec for `react-sketchapp-router`:

- `<SketchRouter>` with `locations` prop (array) of Artboard routes to render.
- `<Switch>` component

Example:

```js
const viewports = [{
  name: 'Mobile', width: 360, height: 640,
}, {
  name: 'Tablet', width: 1024, height: 768,
}];

[...]

Issue #192, opened 2017-10-02
Navigation example


LincMitch:
**Proposal/Feature-request:**

It would be great to see a simple example of basic navigation. For example, 2 pages with next and previous buttons. The pages would be rendered in SketchApp on 2 Artboards, whilst the web render would work as expected.

I think a simple example of this would encourage less tech savvy designers like me to embrace react-sketchapp.

Paarmita:
@LincMitch I would like to work on this issue. What kind of skills are needed to contribute to this?

mathieudutour:
Feel free! You can look at the [react-primitives](https://github.com/airbnb/react-sketchapp/tree/master/examples/profile-cards-primitives) example to see how to render to both sketch and the web. Then it would be designing those 2 pages (nothing fancy, we want the concept to be clear).

Feel free to start a PR and ping me if you are stuck :+1:

Paarmita:
@mathieudutour Is there any other way to install this project if I don't have the sketch?

wellmonge:
@mathieudutour and @Paarmita  I would help you on this issues.

mathieudutour:
Sounds good! Let me know if you need anything

wellmonge:
> Sounds good! Let me know if you need anything

Yes, it would be great if you have any ideia on how to creating prototyping animation, if it's necessary to fit the FR?

mathieudutour:
It’s not supported by react-sketchapp yet so for now we can just render the screen side by side without link between them

PR #483, opened 2020-03-07
Create react-router-prototyping example

macintoshhelper:
Created a `react-sketchapp-router` example – based on a simple blog app. Would this solve #192 (Navigation example) too?

I've created another repo for React Sketch.app examples: https://github.com/elemental-design/react-sketchapp-examples . Any thoughts on it? I'm planning to add a number of examples there like Redux, React hooks and React context API examples there.

**Related issues**
> #343 #192


examples/react-router-prototyping/webpack.skpm.config.js
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @param {object} config - original webpack config.
+ * @param {boolean} isPluginCommand - wether the config is for a plugin command or a resource
+ */
+module.exports = function (config) {
+  /* you can change config here */
+
+  config.resolve = {
+    ...config.resolve,
+    extensions: [...config.resolve.extensions, '.jsx'],
```
mathieudutour: let's just use `.js` instead of `.jsx` like the other examples

examples/react-router-prototyping/package.json
```diff
@@ -0,0 +1,29 @@
...
+  },
+  "author": "Macintosh Helper",
+  "license": "MIT",
+  "devDependencies": {
+    "@skpm/builder": "^0.4.3"
+  },
+  "dependencies": {
+    "prop-types": "^15.5.8",
+    "react": "^16.13.0",
+    "react-router-primitives": "^0.1.2",
```
mathieudutour: I don't think it's used?

macintoshhelper: Do you mean that `prop-types` isn't used? Should I add prop types across the components?

examples/react-router-prototyping/webpack.skpm.config.js
macintoshhelper: I've removed custom webpack config. Is it worth adding a note in the docs that this step is required for `.jsx` files, or is this something that should be addressed in `skpm`?

examples/react-router-prototyping/package.json
mathieudutour: `react-router-primitives` isn't

examples/react-router-prototyping/webpack.skpm.config.js
mathieudutour: we can easily add it to skpm yes.

examples/react-router-prototyping/package.json
macintoshhelper: I'm using it in `examples/react-router-prototyping/src/components/AppBar.js`. My thinking was to use `react-sketchapp-router` in the `App.jsx` entry-point, and `react-router-prototyping` in shared component code.

PR #485, opened 2020-03-25
[fix]docs: display correct emojis on FAQ page

jonathangiamp:
Hi 👋 

I just fixed the display of emojis on FAQ page. As it's a trivial change, I didn't create an issue. Ask me if it's needed 😄


mathieudutour:
Looks good, thanks

PR #160, opened 2017-08-03
feat(inner-shadow): Add possibility to define inner shadows

weaintplastic:
I'd like to be able to define innerShadows. Therefore I've added the following mechanic:

Based on wether a `shadowInner` flag is provided with `styles` the generated shadow will be applied to `style.shadow` or `style.innerShadow`. Also the `_class` is calculated based on the existence of that flag.

Happy to discuss naming and if it makes sense/is possible to apply the same logic to the ImageRenderer.

PR #490, opened 2020-04-13
Add shadowInner style to docs

macintoshhelper:
Add `shadowInner` to `styling.md` docs README.

Code reference: https://github.com/airbnb/react-sketchapp/blob/cee20156d3795c3e4899e746aafde7cd22a149e6/src/jsonUtils/style.ts#L89

**Related issues/PRs**

> #160

### DIFF
--- a/docs/guides/styling.md
+++ b/docs/guides/styling.md
@@ -76,7 +76,7 @@ Components use CSS styles + FlexBox layout.
 | `shadowOffset` | `{ width: number, height: number }` | ✅ |
 | `shadowOpacity` | `number` | ✅ |
 | `shadowRadius` | `number` &#124; `percentage` | ✅ |
-| [see `react-sketchapp` only styles](#react-sketchapp-only-shadow-styles)   | | |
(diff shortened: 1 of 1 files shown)


mathieudutour:
Maybe, because it's a sketch specific style field, it should be in its own section? This section is more like the list of the RN properties that are(n't) supported by react-sketchapp.
What do you think?

macintoshhelper:
👍 Was thinking of that. Should I move `shadowSpread` too, since that's not supported by React Native?

mathieudutour:
Ah yep good catch

macintoshhelper:
I moved the shadows to a new shadow section (instead of layout), to mirror the React Native docs, and added a note of `react-sketchapp` only ones. Or maybe it would make more sense to have a `react-sketchapp` section?

mathieudutour:
Perfect, thanks

macintoshhelper:
Np, thanks for cleaning up the file/text. I had left in extra spacing in the table for stylistic purposes to keep the width aligned for GitHub, but the website GitBook makes the tables full width anyway, so looks good 🙂.

PR #494, opened 2020-04-16
compatible with fontWeight in number

jiazhewang:
The incoming fontWeight is likely to be a number (like `500`) instead of a string (`'500'`), or even other things... [example](https://github.com/ant-design/antd-sketchapp/pull/41)

Maybe it would be safer to avoid these cases in a proactive way.

### DIFF
--- a/src/jsonUtils/sketchImpl/findFontName.ts
+++ b/src/jsonUtils/sketchImpl/findFontName.ts
@@ -94,12 +94,7 @@ export const findFont = (style: TextStyle): NSFont => {
   const defaultFontSize = 14;
 
   const fontSize = style.fontSize ? style.fontSize : defaultFontSize;
-  let fontWeight =
-    style.fontWeight &&
-    String(style.fontWeight).toLowerCase() &&
-    FONT_WEIGHTS[String(style.fontWeight).toLowerCase()]
-      ? FONT_WEIGHTS[String(style.fontWeight).toLowerCase()]
-      : defaultFontWeight;
+  let fontWeight = FONT_WEIGHTS[String(style.fontWeight).toLowerCase()] || defaultFontWeight;
 
   let familyName = defaultFontFamily;
   let isItalic = false;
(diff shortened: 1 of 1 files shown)


src/jsonUtils/sketchImpl/findFontName.ts
```diff
@@ -94,9 +94,12 @@ export const findFont = (style: TextStyle): NSFont => {
   const defaultFontSize = 14;
 
   const fontSize = style.fontSize ? style.fontSize : defaultFontSize;
-  let fontWeight = style.fontWeight
-    ? FONT_WEIGHTS[style.fontWeight.toLowerCase()]
-    : defaultFontWeight;
+  let fontWeight =
+    style.fontWeight &&
+    String(style.fontWeight).toLowerCase() &&
```
mathieudutour, line 99: do we need this check? It doesn't seem necessary.
```suggestion
```

jiazhewang: yea...maybe the whole `fontWeight` could just be:
```suggestion
    let fontWeight = FONT_WEIGHTS[String(style.fontWeight).toLowerCase()] || defaultFontWeight;
```

src/jsonUtils/sketchImpl/findFontName.ts
```diff
@@ -94,9 +94,12 @@ export const findFont = (style: TextStyle): NSFont => {
...
+    FONT_WEIGHTS[String(style.fontWeight).toLowerCase()]
+      ? FONT_WEIGHTS[String(style.fontWeight).toLowerCase()]
+      : defaultFontWeight;
```
jiazhewang, line 102: would this be better?

```suggestion
  let fontWeight = FONT_WEIGHTS[String(style.fontWeight).toLowerCase()] || defaultFontWeight;
```

mathieudutour: yeah looks good

jiazhewang:
Implementation updated.

mathieudutour:
Perfect, thanks!

PR #501, opened 2020-04-19
useWindowDimensions hook for Artboard viewport

### LINKED ISSUE #414: "How to" create UI variations based on screen width
_I am... a designer and a developer

| Requesting a new feature

I can't find any resources on how to alter the rendering of your components based on the the size of the viewport or artboard i.e. single column / two column, show / hide elements.

In Jon Golds' Design as a Function of Data video there is a component called ResponsiveColumnLayout which I presume is doing just this but I cant see a reference to the component in the source code.

A 'how to' on this would be useful or a pointer on an external resource detailing how to do this.

macintoshhelper:
Not sure if it's worth having context feature detection? Or just require people to run recent React versions, rather than have the hook act as a stub on earlier versions.

fixes #414


src/context.tsx
```diff
@@ -0,0 +1,50 @@
...
+    width: PropTypes.number,
+    height: PropTypes.number,
+    name: PropTypes.string,
+  }),
+  children: PropTypes.node,
+};
+
+type ArtboardProps = PropTypes.InferProps<typeof ArtboardPropTypes>;
+
+export const ArtboardProvider = ({ children, viewport }: ArtboardProps & { children: any }) => {
```
mathieudutour, line 34: So I'm not sure it's enough to only consider the viewport. Most of the time people won't specify it, and only use the style's width (like in the basic example https://github.com/airbnb/react-sketchapp/blob/master/examples/basic-setup/src/my-command.js#L45-L51).

So IMO it'd be useful to fallback to the style dimension if we have no viewport (even if we have only the width because it's probably the main use case).

src/context.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const { width, height } = state || {};
+
+  return { width, height };
+}
+
+const ArtboardPropTypes = {
+  viewport: PropTypes.shape({
+    width: PropTypes.number,
+    height: PropTypes.number,
+    name: PropTypes.string,
```
mathieudutour, line 27: Do you think it'd be useful to add `scale` and `fontScale` here? Even if it doesn't change the sketch output, it could be a nice way to provide a hook for people to change those values and control the context

macintoshhelper: Yes, would be great to pass a custom `scale` and `fontScale` to artboards, to better replicate React Native device-specific behaviour. Do you think this should be added to the viewport prop, rather than create a new prop? Not sure if it could create confusion if people think that the `scale`/`fontScale` is supposed to be handled by Sketch viewports?

mathieudutour: It feels natural in `viewport` and I guess with some docs it would be ok.

macintoshhelper: I've filtered out `scale` and `fontScale` from `ArtboardRenderer.tsx` so they don't pollute the generated file. Does this look good/readable?

https://github.com/airbnb/react-sketchapp/blob/aec1caa61bb0302c34f432b9ca42f6b531efca5b/src/renderers/ArtboardRenderer.ts#L25-L35

src/context.tsx
macintoshhelper: Good point, am falling back to `style.width` and `style.height`. Is it worth doing a `console.warn` in the hook, if the `viewport` is missing, prompting them to use the `<Artboard viewport={}` prop? Incase they get confused about why the height is the full page height rather than viewport height?

mathieudutour: Nah I think that what you actually expect: just the size of the artboard. I actually didn't even think about the viewport prop before your PR.

src/context.tsx
mathieudutour: why not just
```js
presetDictionary: { 
     allowResizedMatching: 0, 
     offersLandscapeVariant: 1, 

     name: props.viewport.name, 
     width: props.viewport.width, 
     height: props.viewport.height,      
},
```
?

docs/API.md
```diff
@@ -382,6 +386,39 @@ View primitives
...
+        <Text>Hello World</Text>
+      </View>
+      <View style={{ height, width, backgroundColor: 'blue' }} />
+    </View>
+  );
+};
+
+render(
+  <Page>
+    <Artboard viewport={{ width: 360, height: 640 }}>
```
mathieudutour, line 415: it could be interesting to add 2 Artboard with different viewports and show what `useWindowDimensions` will be in both cases (as a comment below it) - and perhaps showing an additional div in case the width is bigger than a certain breakpoint

src/components/Artboard.tsx
```diff
@@ -28,15 +32,18 @@ export default class Artboard extends React.Component<Props> {
   };
 
   render() {
+    const style = StyleSheet.flatten(this.props.style as UserStyles);
```
mathieudutour, line 35: do you need this type cast here? It's a bit dangerous

macintoshhelper: I'm getting TypeScript errors without it (and in a number of other components too). I added it to get it to build.

mathieudutour: fixed it here: https://github.com/airbnb/react-sketchapp/pull/502

docs/API.md
macintoshhelper: I've added a `devices` array and am mapping through that, and created a conditional `View` for tablet/desktop.

I've created a `useBreakpoint` hook that returns a `getStyle` function, like this: `getStyle(['column', 'row'])` – for column on mobile and row on tablet/desktop. Might be worth featuring in a new doc file, probably overkill for `API.md` example though.

mathieudutour:
I'm going to publish it, feel free to tweet about it if you want - it's a really cool feature and could make for a really visual tweet

PR #512, opened 2020-05-08
Local dev module aliasing for examples

### LINKED ISSUE #363: Automate e2e tests for all examples
Keeping all examples working as API changes are made is cumbersome. I believe the major pain point here is knowing what needs to updated. Currently, this requires manually running each example and verifying output.

I believe we would be able to greatly increase our confidence that examples are all working as expected by creating very light e2e tests. As a first pass, I purpose we simply check that content were rendered rather than a `RedBox`.

This should help us mitigate issues like #351 and #360

macintoshhelper:
This PR creates integration tests for all the examples. The idea is to have a new tests folder that uses `skpm` `test-runner` and Webpack aliasing (to resolve dependencies like `chroma-js`, etc), with all of the dependencies of the examples in a single `package.json`.

This also introduces some better tooling for local development and testing, that should make it easy to setup pre-publish/pre-push hooks that can catch bugs. e.g. `LOCAL_DEV` environment variable to use the local source instead of `react-sketchapp` `npm` package for running `examples` locally, while keeping them individually git cloneable.

`e2e` examples are currently broken, so am doing some debugging to try to see what is causing the test runner crashes first. So far the `symbol.tsx` file seems problematic, maybe because of importing `yoga-layout-prebuilt`.

[WIP], will write more here later.

**Related issues**
> #363

### DIFF
--- a/examples/basic-setup/webpack.skpm.config.js
+++ b/examples/basic-setup/webpack.skpm.config.js
@@ -1,11 +1,13 @@
-module.exports = config => {
-  config.node = {
-    console: false,
-    global: false,
-    process: true,
-    __filename: 'mock',
-    __dirname: 'mock',
-    Buffer: false,
-    setImmediate: false,
-  };
+const path = require('path');
+
+module.exports = (config) => {
+  if (process.env.LOCAL_DEV) {
+    config.resolve = {
+      ...config.resolve,
+      alias: {
+        ...config.resolve.alias,
+        'react-sketchapp': path.resolve(__dirname, '../../lib/'), // FIXME: Should be able to do '../../', as that's a close simulation of an 'npm install' and import 'react-sketchapp'
+      },
+    };
+  }
 };
--- a/examples/basic-setup-typescript/src/my-command.tsx
+++ b/examples/basic-setup-typescript/src/my-command.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import sketch from 'sketch';
 import { render, Artboard, Text, View } from 'react-sketchapp';
 import chroma from 'chroma-js';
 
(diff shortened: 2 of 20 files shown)


macintoshhelper:
Great, sure, feel free to merge, should I add the Webpack config to the rest of the examples too?

mathieudutour:
That’d be nice, but tbh I’m only working on that one hahaha

examples/basic-setup/webpack.skpm.config.js
```diff
@@ -1,11 +1,13 @@
...
-  };
+const path = require('path');
+
+module.exports = (config) => {
+  if (process.env.LOCAL_DEV) {
+    config.resolve = {
+      ...config.resolve,
+      alias: {
+        ...config.resolve.alias,
+        'react-sketchapp': path.resolve(__dirname, '../../lib/'), // FIXME: Should be able to do '../../', as that's a close simulation of an 'npm install' and import 'react-sketchapp'
```
macintoshhelper, line 9: Left in this comment to note that `'../../'` should be used instead of `'../../lib/'`, but it is failing due to using the `lib/module` entrypoint from the `package.json.

macintoshhelper:
Should be good to merge. Also added a quick fix for the TypeScript build – was failing for me due to the `context.document` global.

macintoshhelper:
Also, in my testing of the `@skpm/test-runner` tests, I found that running the tests from `sketchtool` was breaking (whenever `yoga-layout-prebuilt`, etc was imported), but worked fine manually importing `node_modules/@skpm/test-runner/test-runner.sketchplugin` into Sketch and running it from the GUI. Not sure if it's a bug with `test-runner`, or with Sketch itself?

Issue #515, opened 2020-05-10
Svg import broken


macintoshhelper:
_I am..._

| -------------------------------------------------------------------------------------------------
| Reporting a bug or issue
| -------------------------------------------------------------------------------------------------

**Expected behavior:**

![image](https://user-images.githubusercontent.com/6757532/81488125-afa9aa80-925c-11ea-9eb8-8b94061e9d6c.png)

**Observed behavior:**

```sh
> basic-svg@1.0.0 postinstall /Users/<redacted>/basic-svg
> npm run build && skpm-link


> basic-svg@1.0.0 build /Users/<redacted>/basic-svg
> skpm-build

[1/2] 🖨  Copied src/manifest.json in 3ms
error Error while building ./my-command.js
./node_modules/node-gyp-build/index.js
Module not found: Error: Can't resolve 'fs' in '/Users/<redacted>/basic-svg/node_modules/node-gyp-build'
resolve 'fs' in '/Users/<redacted>/basic-svg/node_modules/node-gyp-build'
  Parsed request is a module
  using description file: /Users/<redacted>/basic-svg/node_modules/node-gyp-build/package.json (relative path: .)
    Field 'browser' doesn't contain a valid alias configuration
  after using description file: /Users/<redacted>/basic-svg/node_modules/node-gyp-build/package.json (relative path: .)
    resolve as module

```

**How to reproduce:**

```bash
curl https://codeload.github.com/airbnb/react-sketchapp/tar.gz/master | tar -xz --strip=2 react-sketchapp-master/examples/basic-svg
cd basic-svg
```

**Sketch version:**

`63.1`

**Please attach ~screenshots, a zip file of your project, and~ or a link to your github project**

> https://github.com/airbnb/react-sketchapp/tree/master/examples/basic-svg

macintoshhelper:
I believe this is caused by https://github.com/airbnb/react-sketchapp/blob/9f4b38cdfb7848b32ff9c1968b145dabd72739a9/src/components/index.ts#L8

**Proposed fix:**

```js
import Svg from './Svg';
export { Svg };
```

Fixed in 78c317019391eaa6ea0505c09bdb2e7a2c1ccf76, but may be best to revert and apply in a new atomic PR.

[...]

PR #516, opened 2020-05-10
Integration tests for examples

macintoshhelper:
`skpm-test` appears to be broken due to some `sketchtool` run issue (works fine importing the generated tests manually in Sketch GUI.

So for now, I'm creating a monolithic test example that is added to the test pipeline with `skpm-build __tests/examples/src/main.js` (pseudo-code, haven't tested the npm script yet).

This should make it easy to test all examples with a single command.

### DIFF
--- a/__tests__/examples/webpack.skpm.config.js
+++ b/__tests__/examples/webpack.skpm.config.js
@@ -1,0 +1,24 @@
+const path = require('path');
+
+const aliasedModules = [
+  'chroma-js',
+  'ramda',
+  'react-primitives',
+  '@emotion/primitives',
+  '@emotion/core',
+];
+
+module.exports = (config) => {
+  config.resolve = {
+    ...config.resolve,
+    alias: {
+      ...config.resolve.alias,
+      'react-sketchapp': path.resolve(__dirname, '../../'),
+
+      ...aliasedModules.reduce((acc, mod) => {
+        acc[mod] = path.resolve(__dirname, `./node_modules/${mod}`);
+        return acc;
+      }, {}),
+    },
+  };
+};
--- a/__tests__/examples/package.json
+++ b/__tests__/examples/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "example-tests",
+  "version": "1.0.0",
+  "description": "",
+  "skpm": {
+    "main": "example-tests.sketchplugin",
+    "manifest": "src/manifest.json"
+  },
+  "scripts": {
+    "build": "../../node_modules/.bin/skpm-build",
+    "watch": "../../node_modules/.bin/skpm-build --watch",
+    "render": "../../node_modules/.bin/skpm-build --watch --run",
+    "render:once": "../../node_modules/.bin/skpm-build --run",
+    "postinstall": "npm run build && ../../node_modules/.bin/skpm-link"
+  },
+  "author": "Macintosh Helper <github@macintoshhelper.com>",
+  "license": "MIT",
+  "dependencies": {
+    "@emotion/core": "^10.0.28",
+    "@emotion/primitives": "^10.0.30",
+    "chroma-js": "^1.2.2",
+    "emotion-primitives": "^1.0.0-beta.0",
+    "ramda": "^0.27.0",
+    "react-primitives": "^0.8.1"
+  }
+}
--- a/__tests__/examples/src/main.js
+++ b/__tests__/examples/src/main.js
@@ -1,0 +1,67 @@
+import * as React from 'react';
+import sketch from 'sketch';
+import { render, Document, Page } from 'react-sketchapp';
+
+import { Document as BasicSetup } from '../../../examples/basic-setup/src/my-command';
+import { Document as BasicSvg } from '../../../examples/basic-svg/src/my-command';
+import { Document as Colors } from '../../../examples/colors/src/main';
+import { App as Emotion } from '../../../examples/emotion/src/my-command';
+import { Page as FormValidation } from '../../../examples/form-validation/src/main';
+
+import formValidationData from '../../../examples/form-validation/src/data';
+
+const pages = [
+  {
+    component: BasicSetup,
+    name: 'Basic Setup',
+    data: {
+      colors: {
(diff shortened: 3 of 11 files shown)


src/components/index.ts
```diff
@@ -5,5 +5,5 @@ export { Image } from './Image';
 export { RedBox } from './RedBox';
 export { View } from './View';
 export { Text } from './Text';
-import * as Svg from './Svg';
+import Svg from './Svg';
```
macintoshhelper, line 8: Am happy to revert this if requested. Am doing this to fix the `basic-svg` example to unblock.

package.json
```diff
@@ -38,6 +38,7 @@
     "test:unit": "jest --config jest.config.js --no-watchman",
     "test:ci": "npm run test:unit -- --runInBand",
     "test:e2e": "skpm-test",
+    "test:e2e:examples": "npm run build --prefix __tests__/examples/",
```
macintoshhelper, line 41: Maybe `npm run render:once --prefix ...` is better for now?

macintoshhelper:
@mathieudutour I've got e2e tests working. Took a while to debug it: I found that importing `yoga-layout` off this branch was crashing Sketch, but the same version of `yoga-layout` wasn't crashing in a commit snapshot before the TypeScript refactor, so I downgraded `@skpm/test-runner` to `^0.3` and that fixed it.

There's a bit of clean up to do before a merge, but would be ideal to have a review first. I'm using packages that are installed in `__tests__/examples/package.json` for now. If a new example is added, then it has to be added manually to the e2e examples test file, together with the dependencies in `__tests__/examples/package.json`, so maybe this is something that should be mentioned in `CONTRIBUTING.md`?

I'm still running into #515 (in test runner, and a normal `react-sketchapp` project also), which fails the examples test:

```js
 PASS  basic
 PASS  render-context
 PASS  render-in-wrapped-object

 FAIL  examples

  ● examples › should render examples

     React SketchApp cannot render into this layer. You may be trying to render into a layer that does not take children. Try rendering into a LayerGroup, Artboard, or Page.
      at <anonymous> (../../lib/render.js:23:14)
      at <anonymous> (./examples.test.js:72:2)
      at <anonymous> (../../node_modules/promise-polyfill/lib/index.js:79:0)
      at <anonymous> (../../node_modules/@skpm/timers/timeout.js:18:0)

  ✕ should render examples
```

Patching `import * as Svg from './Svg';` to `import Svg from './Svg';` in `src/components/index.ts` fixes it and the e2e tests pass.

__tests__/skpm/examples.test.js
```diff
@@ -0,0 +1,96 @@
...
+import { App as Emotion } from '../../examples/emotion/src/my-command';
+import { Page as FormValidation } from '../../examples/form-validation/src/main';
+
+import formValidationData from '../../examples/form-validation/src/data';
+
+function getDoc(document) {
+  return sketch.getSelectedDocument() || document;
+}
+
+const examplePages = [
```
mathieudutour, line 17: a lot fo this is duplicated in `main.js`. Could easily import it from there

__tests__/skpm/.babelrc
```diff
@@ -0,0 +1,14 @@
+{
```
mathieudutour, line 1: why do you need this file?

__tests__/examples/webpack.skpm.config.js
```diff
@@ -0,0 +1,24 @@
+const path = require('path');
+
+const aliasedModules = [
```
mathieudutour, line 3: could we get this list from the package.json? Just to make sure nothing is missing

package.json
```diff
@@ -80,7 +81,7 @@
   },
   "devDependencies": {
     "@skpm/babel-preset": "^0.2.1",
-    "@skpm/test-runner": "^0.4.1",
+    "@skpm/test-runner": "^0.3.2",
```
mathieudutour, line 84: hum that's unfortunate. I'll try to see what's up

__tests__/skpm/.babelrc
macintoshhelper: I pulled it from before the TypeScript refactor. I believe `@babel/plugin-transform-modules-commonjs` is needed for `@skpm/test-runner`? `@skpm/test-runner@0.4.1` has this plugin, but without `"noInterop": true`.

Without the `.babelrc` file, the tests give this error:

```bash
undefined is not an object (evaluating 'module.exports.tests = __skpm_tests__')
```

package.json
```diff
@@ -30,13 +30,15 @@
     "docs:watch": "npm run docs:prepare && gitbook serve",
     "docs:publish": "npm run docs:clean && npm run docs:build && cd _book && git init && git commit --allow-empty -m 'update book' && git fetch https://github.com/airbnb/react-sketchapp gh-pages && git checkout -b gh-pages && git add . && git commit -am 'update book' && git push https://github.com/airbnb/react-sketchapp gh-pages --force",
     "lint-staged": "lint-staged",
-    "prepublishOnly": "npm run clean && npm run test:ci && npm run build",
+    "prepublishOnly": "npm run clean && npm run test:ci && npm run install:e2e && npm run build",
```
macintoshhelper, line 33: `npm run install:e2e` triggers the `__tests__/examples` `postinstall` script. Not sure if this should be left in, it'll fail the prepublish script if the examples `skpm-build` fails, but not sure if it'll work in CI.

__tests__/examples/webpack.skpm.config.js
```diff
@@ -0,0 +1,20 @@
...
+const pkg = require(path.resolve(__dirname, 'package.json'));
+
+const aliasedModules = Object.keys(pkg.dependencies);
+
+module.exports = (config) => {
+  config.resolve = {
+    ...config.resolve,
+    alias: {
+      ...config.resolve.alias,
+      'react-sketchapp': path.resolve(__dirname, '../../lib'),
```
macintoshhelper, line 12: Seems that `@skpm/test-runner` needs the commonjs build, and doesn't work with the ES build, strange.

PR #520, opened 2020-06-08
fix the SVG component export

mathieudutour:


macintoshhelper:
Great, thanks. I've tested this with the `basic-svg` example and it works fine.

Tested with:
- Destructured usage – `const { Path } = Svg;`
- `react-native-svg` style import – `import Svg, { G, Path } from 'react-sketchapp/lib/components/Svg';`
- Static usage – `Svg.Path`

